PR #1, opened 2019-08-08
Update pom parent and scm details

AnanaMJ:
I've removed `cobertura-maven-plugin` since this is not in `eg-oss-plugin`'s dependencies (it was added for making it work with `hotels-oss-plugin`) :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,15 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -25,9 +25,9 @@
   </modules>
 
   <scm>
-    <connection>scm:git:git@github.com:HotelsDotCom/beekeeper.git</connection>
-	<developerConnection>scm:git:git@github.com:HotelsDotCom/beekeeper.git</developerConnection>
-	<url>https://github.com/HotelsDotCom/beekeeper</url>
+    <connection>scm:git:git@github.com:ExpediaGroup/beekeeper.git</connection>
+    <developerConnection>scm:git:git@github.com:ExpediaGroup/beekeeper.git</developerConnection>
+    <url>https://github.com/HotelsDotCom/beekeeper</url>
```
patduin: still says hotels

AnanaMJ: Ah, I missed one! Thanks :D (it's gone now)

pom.xml
```diff
@@ -1,15 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
max-jacobs: Looks like this is now on two lines, do we want to be changing this?

AnanaMJ: I would think so; it was over 120 characters long :)

PR #2, opened 2019-08-22
Catching multiobject exception

max-jacobs:
Adding catch for MutliObjectException

### DIFF
--- a/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupService.java
+++ b/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupService.java
@@ -30,6 +30,8 @@ import org.springframework.data.domain.Pageable;
 
 import io.micrometer.core.annotation.Timed;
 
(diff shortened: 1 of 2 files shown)


beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupService.java
```diff
@@ -88,6 +90,15 @@ private void cleanupContent(EntityHousekeepingPath housekeepingPath) {
       log.info("Cleaning up path \"{}\"", housekeepingPath.getPath());
       pathCleaner.cleanupPath(housekeepingPath.getPath(), housekeepingPath.getTableName());
       updateAttemptsAndStatus(housekeepingPath, PathStatus.DELETED);
+    } catch (MultiObjectDeleteException e) {
```
patduin: I don't think this is the place to do it. I think it should be done in the S3PathCleaner. That's the class that knows about AWS and this is an AWS exception that should bubble up. If we change the implementation of PathCleaner this catch will be useless.

PR #3, opened 2019-08-27
Open source docker build

max-jacobs:
Updating build to only produce one, versioned image per application and not push. Updating Jenkinsfile to push images on deploys and tag as latest on release. This presumes that snapshot/release images will be in the same repo.

Still required:
- `eg-oss-ci` user
- `eg-oss-settings`
- `docker.hub` property in `eg-oss-parent`
- Test that `DOCKER_ORG = readMavenPom().properties['docker.org']` works as expected

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -66,7 +66,7 @@ pipeline {
                   -Dresume=false \
                   -DreleaseVersion=${RELEASE_VERSION} \
                   -DdevelopmentVersion=${DEVELOPMENT_VERSION} \
-                  -DautoVersionSubmodules=true"""{
+                  -DautoVersionSubmodules=true"""
         }
         echo 'Pushing images...'
         docker tag $DOCKER_ORG/beekeeper-cleanup:${RELEASE_VERSION} $DOCKER_ORG/beekeeper-cleanup:latest
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,8 +49,8 @@ pipeline {
         withCredentials([usernamePassword(credentialsId: 'dockerhub-egopensource', passwordVariable: 'PASSWORD', usernameVariable: 'USERNAME')]) {
           docker login -u $USERNAME -p $PASSWORD
         }
-        // docker push $DOCKER_ORG/beekeeper-cleanup
-        // docker push $DOCKER_ORG/beekeeper-path-scheduler-apiary
+        // docker push ${DOCKER_ORG}/beekeeper-cleanup
+        // docker push ${DOCKER_ORG}/beekeeper-path-scheduler-apiary
       }
     }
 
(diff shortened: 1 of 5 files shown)


beekeeper-package/pom.xml
```diff
@@ -17,28 +17,9 @@
...
-          <value>true</value>
-        </property>
-      </activation>
-      <properties>
-        <docker.repo>${artifactory.docker.release.repository.id}.${artifactory.base}</docker.repo>
-      </properties>
-    </profile>
-  </profiles>
+  <properties>
+    <docker.org></docker.org>
```
AnanaMJ: Should this be something other than null?

Jenkinsfile
```diff
@@ -20,15 +22,26 @@ pipeline {
         echo 'Checking out project...'
         checkout scm
         echo 'Building...'
-        withMaven(jdk: 'OpenJDK11', maven: 'Maven3.6', mavenSettingsConfig: 'hcomdata-artifactory-maven-settings') {
-          sh 'mvn clean deploy jacoco:report checkstyle:checkstyle'
+        withMaven(jdk: 'OpenJDK11', maven: 'Maven3.6', mavenSettingsConfig: 'hcomdata-oss-settings') {
```
AnanaMJ: Is `hcomdata-oss-settings` something already created in Jenkins?

And should we make one for eg-oss?

beekeeper-package/pom.xml
max-jacobs: Removed from here, this will be in eg-oss-parent

Jenkinsfile
```diff
@@ -9,9 +9,11 @@ pipeline {
   }
 
   environment {
-    GIT = credentials('s-hcom-jenkins-rbac')
+    GIT = credentials('s-hcom-ci')
     GIT_USERNAME = "${env.GIT_USR}"
     GIT_PASSWORD = "${env.GIT_PSW}"
+
+    DOCKER_ORG = readMavenPom().getProperties().getProperty('docker.org')
```
AnanaMJ: You're adding `docker.org` to the parent pom, right?

Jenkinsfile
```diff
@@ -48,13 +61,18 @@ pipeline {
...
+        withMaven(jdk: 'OpenJDK11', maven: 'Maven3.6', mavenSettingsConfig: 'hcomdata-oss-settings') {
           sh """mvn --batch-mode release:prepare release:perform \
                   -Dresume=false \
                   -DreleaseVersion=${RELEASE_VERSION} \
                   -DdevelopmentVersion=${DEVELOPMENT_VERSION} \
                   -DautoVersionSubmodules=true"""
         }
+        echo 'Pushing images...'
+        docker tag $DOCKER_ORG/beekeeper-cleanup:${RELEASE_VERSION} $DOCKER_ORG/beekeeper-cleanup:latest
+        docker tag $DOCKER_ORG/beekeeper-path-scheduler-apiary:${RELEASE_VERSION} $DOCKER_ORG/beekeeper-cleanup:latest
```
AnanaMJ: it should be `$DOCKER_ORG/beekeeper-path-scheduler-apiary:latest`

beekeeper-package/pom.xml
AnanaMJ: Ohhh okay :)

Jenkinsfile
max-jacobs: We have a fresh project there, so its empty, but I think you're right we'll need to use different settings for eg open source. Good spot, thanks

max-jacobs:
Closing this PR to do some testing in Jenkins

PR #4, opened 2019-09-02
Oss docker build

max-jacobs:
Reopening this PR for the open source build.

Will squash when merging so we don't have all these testing commits.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,9 +15,8 @@ pipeline {
 
     DOCKER_ORG = readMavenPom().getProperties().getProperty('docker.org')
 
-    withCredentials([file(credentialsId: 'eg-oss-settings.xml', variable: 'SETTINGS')]) {
-      MAVEN_SETTINGS = SETTINGS
-    }
+    MAVEN_SETTINGS = credentials('eg-oss-settings.xml')
+
     OSS_GPG_PUB_KEYRING = credentials('pubring.gpg')
     OSS_GPG_SEC_KEYRING = credentials('secring.gpg')
     OSS_GPG_PASSPHRASE = credentials('private-key-passphrase')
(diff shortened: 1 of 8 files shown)


Jenkinsfile
```diff
@@ -2,16 +2,28 @@ pipeline {
...
     GIT_PASSWORD = "${env.GIT_PSW}"
+
+    DOCKER_REGISTRY = 'expediagroup'
+
+    MVN_OSS_SETTINGS = credentials('eg-oss-settings.xml')
+
+    OSS_GPG_PUB_KEYRING = credentials('pubring.gpg')
+    OSS_GPG_SEC_KEYRING = credentials('secring.gpg')
+    OSS_GPG_KEYNAME='oss@expediagroup.com'
+    OSS_GPG_PASSPHRASE = credentials('private-key-passphrase')
```
AnanaMJ: I'll assume there are correct 😆

Jenkinsfile
```diff
@@ -20,15 +32,24 @@ pipeline {
         echo 'Checking out project...'
         checkout scm
         echo 'Building...'
-        withMaven(jdk: 'OpenJDK11', maven: 'Maven3.6', mavenSettingsConfig: 'hcomdata-artifactory-maven-settings') {
-          sh 'mvn clean deploy jacoco:report checkstyle:checkstyle'
+        sh 'mvn clean deploy jacoco:report checkstyle:checkstyle -s ${MVN_OSS_SETTINGS}'
```
massdosage: We should add spotbugs in here (can be a separate PR)

max-jacobs: I did try this, but got a maven error saying it couldn't be found. Will make this a separate PR.

massdosage: Yes, you need to use the latest oss-parent (release this morning!) to get it.

PR #6, opened 2019-09-03
Removing old test

max-jacobs:
I can't get this test to fail locally, but it has broken the Jenkins build twice consecutively.

Seeing as we aren't using gauge metrics anymore, we might as well remove it.

I think what's happening is that both paths are deleted and so the metric reported is `CONTENT.getBytes().length + doubleContent.getBytes().length` which fails on the assert at line 253. Perhaps our machines are fast enough in this loop to catch the metric before the second path is deleted.

Any thoughts?


max-jacobs:
> I think the test made sense when we used the gauge, but now we can just check that the final value reported is the sum of the sizes, which is already done, right?

I think we are probably covered by `S3BytesDeletedReporterTest`

PR #13, opened 2019-10-22
Add health endpoint

max-jacobs:
PR to add health check endpoints to Beekeeper

Endpoints will be at Spring's default location: `/actuator/health`

### DIFF
--- a/beekeeper-integration-tests/pom.xml
+++ b/beekeeper-integration-tests/pom.xml
@@ -19,31 +19,11 @@
       <groupId>com.expediagroup</groupId>
       <artifactId>beekeeper-path-scheduler-apiary</artifactId>
       <version>${project.version}</version>
-      <exclusions>
-        <exclusion>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-web</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>org.springframework.boot</groupId>
-          <artifactId>spring-boot-starter-actuator</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.expediagroup</groupId>
       <artifactId>beekeeper-cleanup</artifactId>
       <version>${project.version}</version>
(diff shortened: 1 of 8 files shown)


beekeeper-integration-tests/pom.xml
```diff
@@ -19,11 +19,31 @@
       <groupId>com.expediagroup</groupId>
       <artifactId>beekeeper-path-scheduler-apiary</artifactId>
       <version>${project.version}</version>
+      <exclusions>
```
patduin: this is weird though and I would think the same problem occurs when you run it unless something the Integration tests pull in that doesn't have this. or do we repackage the jars or something?
Would be nice to have a test for the healthcheck :/

max-jacobs: Fixed, now excluding near the source in the path scheduler. Also I've added integration tests for the healthcheck.

I had to change `assertMertrics()` as we used to assert that the list of metrics was a certain size, but the web dependency adds a whole bunch of extra metrics for free (~90 or so).

beekeeper-path-scheduler-apiary/pom.xml
```diff
@@ -62,6 +62,18 @@
           <groupId>org.apache.logging.log4j</groupId>
           <artifactId>log4j-slf4j-impl</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>javax.servlet</groupId>
```
max-jacobs: Removed `servlet-api` and `javax.servlet` as the ITs wouldnt start (though the applications would).

Removed `geronimo-jaspic_1.0_spec` because calling the endpoint `/actuator/health` during the ITs threw this error (though was fine when running the applications):

```
java.lang.SecurityException: AuthConfigFactory error: java.lang.ClassNotFoundException: org.apache.geronimo.components.jaspi.AuthConfigFactoryImpl
```

did some Googling, didn't find much. But removing this package meant the tomcat equivalent `AuthConfigFactoryImpl` was used, and that runs fine.

PR #14, opened 2019-11-04
Changing MeterRegistry to GraphiteMeterRegistry

max-jacobs:
Currently, we create a `MeterRegistry` bean which has the config for graphite metrics. Because its not explicitly a `GraphiteMeterRegistry` bean, Spring creates its own `GraphiteMeterRegistry` bean to publish its default web metrics. So even though `graphite.enabled: false` in the config, there will be errors in the logs as the host and port isn't configured properly for the Spring default registry.

Relevant Spring configuration files:
- [GraphiteMetricsExportAutoConfiguration.java](https://github.com/spring-projects/spring-boot/blob/07ca7747fa7c8e16f75ea99fa44d38c675dff780/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteMetricsExportAutoConfiguration.java#L60)
- [GraphiteProperties.java](https://github.com/spring-projects/spring-boot/blob/07ca7747fa7c8e16f75ea99fa44d38c675dff780/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteProperties.java#L34)

This PR changes this so that now the default Spring metrics should use the graphite config that we provide. This matches how Waggle Dance configures this bean (see [here](https://github.com/HotelsDotCom/waggle-dance/blob/586bf4df813efc72e043fde9282eea7ccdb96b09/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/metrics/MonitoringConfiguration.java#L50)).

I've tested this locally and when setting `graphite.enabled: false`, there are no longer any errors. When true our metrics are still published _and_ we now get the spring web metrics for free.

### DIFF
--- a/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedReporter.java
+++ b/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedReporter.java
@@ -23,7 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.micrometer.core.instrument.Counter;
-import io.micrometer.graphite.GraphiteMeterRegistry;
+import io.micrometer.core.instrument.MeterRegistry;
 
 import com.amazonaws.services.s3.model.DeleteObjectsRequest;
 
@@ -37,7 +37,7 @@ public class S3BytesDeletedReporter {
   private final Map<String, Long> keyToSize = new HashMap<>();
   private final Counter counter;
 
-  public S3BytesDeletedReporter(S3Client s3Client, GraphiteMeterRegistry meterRegistry, boolean dryRunEnabled) {
+  public S3BytesDeletedReporter(S3Client s3Client, MeterRegistry meterRegistry, boolean dryRunEnabled) {
     this.s3Client = s3Client;
 
     String metricName = dryRunEnabled ? DRY_RUN_METRIC_NAME : METRIC_NAME;
--- a/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedReporter.java
+++ b/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedReporter.java
@@ -23,7 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.micrometer.core.instrument.Counter;
-import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.graphite.GraphiteMeterRegistry;
 
 import com.amazonaws.services.s3.model.DeleteObjectsRequest;
 
@@ -37,7 +37,7 @@ public class S3BytesDeletedReporter {
   private final Map<String, Long> keyToSize = new HashMap<>();
   private final Counter counter;
 
-  public S3BytesDeletedReporter(S3Client s3Client, MeterRegistry meterRegistry, boolean dryRunEnabled) {
+  public S3BytesDeletedReporter(S3Client s3Client, GraphiteMeterRegistry meterRegistry, boolean dryRunEnabled) {
     this.s3Client = s3Client;
 
     String metricName = dryRunEnabled ? DRY_RUN_METRIC_NAME : METRIC_NAME;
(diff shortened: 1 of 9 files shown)


beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/context/MonitoringContext.java
```diff
@@ -32,7 +31,7 @@
 public class MonitoringContext {
 
   @Bean
-  MeterRegistry registry(GraphiteConfigFactory graphiteConfigFactory) {
+  GraphiteMeterRegistry registry(GraphiteConfigFactory graphiteConfigFactory) {
```
massdosage: I remember us having to do something similar in WaggleDance for the same reason (see https://github.com/HotelsDotCom/waggle-dance/blob/master/waggle-dance-core/src/main/java/com/hotels/bdp/waggledance/metrics/MonitoringConfiguration.java). There we also renamed the method to be `graphiteMeterRegistry` - not sure what (if any) difference that makes.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedReporter.java
```diff
@@ -37,7 +37,7 @@
   private final Map<String, Long> keyToSize = new HashMap<>();
   private final Counter counter;
 
-  public S3BytesDeletedReporter(S3Client s3Client, MeterRegistry meterRegistry, boolean dryRunEnabled) {
+  public S3BytesDeletedReporter(S3Client s3Client, GraphiteMeterRegistry meterRegistry, boolean dryRunEnabled) {
```
massdosage: Just checking, these don't work with the more generic type? i.e. if you make the change in the beans to be `GraphiteMeterRegistry` can't these stay as just `MeterRegistry` in the classes which use them? I can see how in the config you want to go for a specific type but if you make this change here then these classes can't be used with any other type of `MeterRegistry` which feels like a shame to introduce such coupling.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/context/MonitoringContext.java
max-jacobs: Yep spotted that in WD. Not sure if it makes a difference, but will rename for consistency.

max-jacobs: It does make a (small) difference; the method name affects the bean name, which is used somewhere else. All good.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedReporter.java
max-jacobs: Yep good point, I don't think these necessarily have to change. Will revert them back. Done.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/BeekeeperPathSchedulerApiary.java
```diff
@@ -52,7 +52,7 @@ public static void main(String[] args) {
   public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
     context = (ConfigurableApplicationContext) applicationContext;
     runner = (PathSchedulerApiaryRunner) context.getBean("pathSchedulerApiaryRunner");
-    meterRegistry = (MeterRegistry) context.getBean("registry");
+    meterRegistry = (MeterRegistry) context.getBean("graphiteMeterRegistry");
```
massdosage: Ah, OK, so that's what had to change. Up to you whether you prefer the method named "registry" or this. The former is more generic downstream - especially since you're casting it to the parent class here so theoretically you could change the type in the Context and everything would still work here as long as the method was still called "registry". If we name the method after the type it's more "correct" in the Context but does mean if we change the type there and give it a more descriptive method name then we need to change it here too.

max-jacobs: Will go for the compromise of "meterRegistry" -> that's probably the most descriptive.

massdosage: Even better yes!

PR #17, opened 2019-11-18
Using default cleanup delay if there is a problem with the table setting

max-jacobs:


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.1.5] - TBD
+### Changed
+- Using default deletion delay if table parameter is configured incorrectly
```
massdosage: Fuuuuuullll stop!

max-jacobs: .!

max-jacobs:
@massdosage 

We have a ticket to update the metrics that Beekeeper reports, so I can add this issue to that ticket? That metrics ticket will include a few improvements so would probably be better there.

Will update the readme

massdosage:
> @massdosage
> 
> We have a ticket to update the metrics that Beekeeper reports, so I can add this issue to that ticket? That metrics ticket will include a few improvements so would probably be better there.
> 
> Will update the readme

OK, yes, please add this to the AC on that ticket and then mention it here (you might need to create a GH version of it?)

max-jacobs:
Merging this and creating new feature here: #18

PR #21, opened 2019-11-26
Handle hadoop s3 uris

max-jacobs:
If the path starts with "s3", this change will ensure that the scheme used for the S3 URI will be "s3" only and not "s3a" or "s3n".

### DIFF
--- a/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
+++ b/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
@@ -1,0 +1,48 @@
+/**
+ * Copyright (C) 2019 Expedia, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
```diff
@@ -0,0 +1,48 @@
...
+
+  private AmazonS3URI amazonS3URI;
+
+  public S3SchemeURI(String housekeepingPath) throws URISyntaxException {
+    if (!housekeepingPath.startsWith("s3")) {
+      throw new BeekeeperException(format("'%s' is not an S3 path.", housekeepingPath));
+    }
+    URI uri = URI.create(housekeepingPath);
+    URI s3Uri = new URI("s3", uri.getHost(), uri.getPath(), null);
+    this.amazonS3URI = new AmazonS3URI(s3Uri);
```
patduin: We were doing new AmazonS3URI(s3Uri, false) before (false==uriEncode) any reason why we are not doing that now? Is it worth testing this?

max-jacobs: From the AmazonS3URI docs: 

> URL encoding is recommended if you work with bucket names or object keys with special characters.

Which I don't _think_ we are. But anyway, given that it was false before, we weren't using this feature of `AmazonS3URI`, so theres no change here.

PR #22, opened 2019-11-26
Handle hadoop s3 uris

max-jacobs:
The original housekeepingPath was being used to clean up the sentinel files (so could still be "s3n" or "s3a"), this change uses the path which we are sure is "s3".

### DIFF
--- a/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3PathCleaner.java
+++ b/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3PathCleaner.java
@@ -67,7 +67,7 @@ public class S3PathCleaner implements PathCleaner {
       try {
         String path = s3SchemeURI.getPath();
         if (s3SchemeURI.getPath().endsWith("/")) {
-          path = path.substring(0, housekeepingPath.length() - 1);
+          path = path.substring(0, path.length() - 1);
         }
         sentinelFilesCleaner.deleteSentinelFiles(path);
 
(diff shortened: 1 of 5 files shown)


beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3PathCleaner.java
```diff
@@ -65,14 +65,15 @@ public void cleanupPath(String housekeepingPath, String tableName) {
       deleteFilesInDirectory(bucket, key);
 
       try {
-        if (housekeepingPath.endsWith("/")) {
-          housekeepingPath = housekeepingPath.substring(0, housekeepingPath.length() - 1);
+        String path = s3SchemeURI.getPath();
+        if (s3SchemeURI.getPath().endsWith("/")) {
+          path = path.substring(0, housekeepingPath.length() - 1);
```
patduin: is this correct? `housekeepingPath.length() - 1`

max-jacobs: Good spot I missed one :(

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3PathCleaner.java
```diff
@@ -65,14 +65,15 @@ public void cleanupPath(String housekeepingPath, String tableName) {
       deleteFilesInDirectory(bucket, key);
 
       try {
-        if (housekeepingPath.endsWith("/")) {
-          housekeepingPath = housekeepingPath.substring(0, housekeepingPath.length() - 1);
+        String path = s3SchemeURI.getPath();
+        if (s3SchemeURI.getPath().endsWith("/")) {
```
JayGreeeen: Why not just have `path.endsWith("/")` since you already defined it above?

PR #24, opened 2019-12-03
Table level timer

max-jacobs:
This PR creates a new annotation and aspect so that we can start recording table level metrics:

- `@TimedHousekeepingPath` annotation to time `HousekeepingPath` methods
- `TimedHousekeepingPathAspect` to be triggered by the annotation and do the timing.

I've had to change the signature of some of the methods being timed already and update the tests accordingly. I've also added tests for the new annotation and aspect.

### DIFF
--- a/beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/path/aws/S3DryRunPathCleanerTest.java
+++ b/beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/path/aws/S3DryRunPathCleanerTest.java
@@ -17,9 +17,9 @@ package com.expediagroup.beekeeper.cleanup.path.aws;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
-import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+
+import java.time.Duration;
+import java.time.LocalDateTime;
 
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
@@ -29,10 +29,11 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 import io.findify.s3mock.S3Mock;
-import io.micrometer.core.instrument.Counter;
-import io.micrometer.core.instrument.MeterRegistry;
 
 import com.amazonaws.services.s3.AmazonS3;
+
+import com.expediagroup.beekeeper.cleanup.monitoring.BytesDeletedReporter;
+import com.expediagroup.beekeeper.core.model.EntityHousekeepingPath;
 
 @ExtendWith(MockitoExtension.class)
 class S3DryRunPathCleanerTest {
@@ -45,23 +46,29 @@ class S3DryRunPathCleanerTest {
   private final String partition1Sentinel = "table/id1/partition_1_$folder$";
   private final String absolutePath = "s3://" + bucket + "/" + keyRoot;
   private final String tableName = "table";
+  private final String databaseName = "database";
 
   private final S3Mock s3Mock = new S3Mock.Builder().withPort(0).withInMemoryBackend().build();
+  private EntityHousekeepingPath housekeepingPath;
   private AmazonS3 amazonS3;
   private S3Client s3Client;
-  private S3BytesDeletedReporter s3BytesDeletedReporter;
-  private @Mock MeterRegistry meterRegistry;
+  private @Mock BytesDeletedReporter bytesDeletedReporter;
 
   private S3PathCleaner s3DryRunPathCleaner;
 
   @BeforeEach
   void setUp() {
-    when(meterRegistry.counter(anyString())).thenReturn(mock(Counter.class));
     amazonS3 = AmazonS3Factory.newInstance(s3Mock);
     amazonS3.createBucket(bucket);
     s3Client = new S3Client(amazonS3, true);
-    s3BytesDeletedReporter = new S3BytesDeletedReporter(s3Client, meterRegistry, false);
-    s3DryRunPathCleaner = new S3PathCleaner(s3Client, new S3SentinelFilesCleaner(s3Client), s3BytesDeletedReporter);
+    s3DryRunPathCleaner = new S3PathCleaner(s3Client, new S3SentinelFilesCleaner(s3Client), bytesDeletedReporter);
+    housekeepingPath = new EntityHousekeepingPath.Builder()
+      .path(absolutePath)
+      .tableName(tableName)
+      .databaseName(databaseName)
+      .creationTimestamp(LocalDateTime.now())
+      .cleanupDelay(Duration.ofDays(1))
+      .build();
   }
 
   @AfterEach
@@ -74,7 +81,7 @@ class S3DryRunPathCleanerTest {
     amazonS3.putObject(bucket, key1, content);
     amazonS3.putObject(bucket, key2, content);
 
-    s3DryRunPathCleaner.cleanupPath(absolutePath, tableName);
+    s3DryRunPathCleaner.cleanupPath(housekeepingPath);
 
     assertThat(amazonS3.doesObjectExist(bucket, key1)).isTrue();
     assertThat(amazonS3.doesObjectExist(bucket, key2)).isTrue();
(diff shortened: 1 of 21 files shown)


beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/monitoring/TimedHousekeepingPathAspect.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Aspect for intercepting methods annotated with {@link TimedHousekeepingPath}.
+ *
+ * The method can have any number of arguments but a {@link HousekeepingPath} must be the first. Will add the fully
+ * qualified table name to the timer metric as a tag.
+ */
+@Aspect
+@Component
+public class TimedHousekeepingPathAspect {
```
jmnunezizu: I don't see the implementation of the previous `Timed` aspect, but this one seems to be too specific and could be made a bit more generic.

The way I see it, the only difference is that it requires the object to contain the `tableName`.

I see two options:

a. Move the method to an interface that returns the tableName and change this aspect to a `TimedTableAspect`; or
b. Make the class implement a MetricTag interface and delegate the responsibility of returning what bit will be used to make the full metricTags to the implementing class.

The way it is right now, I see a potential of having to create very similar (but different) aspects for other metrics.

max-jacobs: b. sounds delicious, will give it a go to decouple this annotation from the `HousekeepingPath`

pom.xml
```diff
@@ -115,6 +115,22 @@
...
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>com.mycila</groupId>
+          <artifactId>license-maven-plugin</artifactId>
+          <version>${license.maven.plugin.version}</version>
+          <configuration>
+            <!-- excluding files that don't need a header update -->
+            <excludes>
+              <exclude>src/main/java/com/expediagroup/beekeeper/core/monitoring/TimedHousekeepingPathAspect.java</exclude>
```
nvitucci: Why is this the only file to be explicitly excluded?

nvitucci: Had a chat with Max and clarified this.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/monitoring/TimedHousekeepingPathAspect.java
max-jacobs: This is now done. I've created a `Taggable` interface with one method `getMetricTag()`, which returns a `MetricTag` with the key for a tag and the tag itself. More decoupled/reusable now. Let me know what you think, cheers!

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/path/aws/S3DryRunPathCleanerTest.java
```diff
@@ -45,23 +45,29 @@
...
 
   @BeforeEach
   void setUp() {
-    when(meterRegistry.counter(anyString())).thenReturn(mock(Counter.class));
     amazonS3 = AmazonS3Factory.newInstance(s3Mock);
     amazonS3.createBucket(bucket);
     s3Client = new S3Client(amazonS3, true);
-    s3BytesDeletedReporter = new S3BytesDeletedReporter(s3Client, meterRegistry, false);
     s3DryRunPathCleaner = new S3PathCleaner(s3Client, new S3SentinelFilesCleaner(s3Client), s3BytesDeletedReporter);
+    housekeepingPath = new EntityHousekeepingPath.Builder()
```
jmnunezizu: This is not that important, but I quite like when I can use the Builder pattern combined with a static factory method.

```java
housekeepingPath = EntityHousekeepingPath.builder()
  .path(...)
  ...
  .build();
```

The `builder` method returns a new instance of `Builder`.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/monitoring/TimedTaggableAspect.java
```diff
@@ -0,0 +1,72 @@
...
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+
+/**
```
jmnunezizu: This is quite nice now. Much more reusable. Nice!

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/PathCleaner.java
```diff
@@ -15,7 +15,10 @@
  */
 package com.expediagroup.beekeeper.cleanup.path;
 
+import com.expediagroup.beekeeper.core.model.HousekeepingPath;
+
 public interface PathCleaner {
 
-  void cleanupPath(String housekeepingPath, String tableName);
+  void cleanupPath(HousekeepingPath housekeepingPath);
```
patduin: We don't publish this as an API right, cause this is backward incompatible if just used internally we're good

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
```diff
@@ -122,8 +117,10 @@ void typicalOnePageWithMultipleElements() {
 
     pagingCleanupService.cleanUp(Instant.now());
 
-    verify(pathCleaner).cleanupPath(entityPath1.getPath(), tableName);
-    verify(pathCleaner).cleanupPath(entityPath2.getPath(), tableName);
+    verify(pathCleaner, times(2)).cleanupPath(pathCaptor.capture());
+    assertThat(pathCaptor.getAllValues())
+      .extracting("path")
+      .containsExactlyInAnyOrder(path1, path2);
```
patduin: the order is not relevant anymore?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/PathCleaner.java
max-jacobs: The only thing I can think of which would be affected is the vacuum tool, which I'll check now

max-jacobs: This is fine, the vacuum tool doesn't use the cleaner, just the scheduler which hasn't changed

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
max-jacobs: Good spot, will use `containsExactly()`

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/path/aws/S3DryRunPathCleanerTest.java
max-jacobs: >  Builder pattern combined with a static factory method.

Not sure what you mean by this, what would this look like? thanks!

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/PathCleaner.java
patduin: ok

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/path/aws/S3DryRunPathCleanerTest.java
jmnunezizu: Discussed in person. Now the example makes sense.

max-jacobs: I see what you mean, will add this to my list of useful refactorings :) thanks

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/monitoring/TimedTaggableAspect.java
```diff
@@ -0,0 +1,72 @@
...
```
massdosage: This still needs a license header - it would be copyright Expedia and Pivotal and then under the Apache license.

max-jacobs: Oh really? Thanks for the heads up, shall I do this manually then? Or is there a way to add this to the mycila plugin?

Will create a separate PR for this file, cheers.

massdosage: Yeah, simplest is to do it manually and then tell Mycila not to overwrite it (that bit you've already done).

PR #25, opened 2019-12-10
Table level bytes deleted

max-jacobs:
PR to refactor and update deletions so that we can report table level bytes deleted.

- Separated the `S3BytesDeletedReporter` in to a "calculator" which remembers what files/bytes are deleted, and a more generic reporter which reports the metric
- Refactored the `S3PathCleaner` slightly to make it a little easier to follow - please let me know if you disagree/have any ideas for improvements!
- new tests etc
- a new `FileSystem` enum (at the moment only S3 is supported) - i'm not 100% sure about this so feedback is welcome.

Let me know if you have any questions.

Thanks!


beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporter.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  public void report(long bytesDeleted, String fullyQualifiedTableName, FileSystem fileSystem) {
+    String fileSystemMetricName = String.join("-", fileSystem.toString()
+      .toLowerCase(), metricName);
+    Counter counter = Search.in(meterRegistry)
+      .name(fileSystemMetricName)
+      .tags(tags(fullyQualifiedTableName))
+      .counter();
+    if (counter == null) {
```
massdosage: Would this method ever be called by multiple threads? If so what would the effect be of registering the counter multiple times? My guess is that the last thread to exit would override the first one and we'd lose the first thread's metrics...

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedCalculator.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private S3Client s3Client;
+  private Map<String, Long> keyToSize = new HashMap<>();
+  private long bytesDeleted = 0;
+
+  public S3BytesDeletedCalculator(S3Client s3Client) {
+    this.s3Client = s3Client;
+  }
+
+  public void cacheFiles(String bucket, List<String> keys) {
```
massdosage: `cacheFileSizes` maybe more descriptive?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3PathCleaner.java
```diff
@@ -53,30 +56,22 @@ public void cleanupPath(HousekeepingPath housekeepingPath) {
...
-      } catch (Exception e) {
-        log.warn("Sentinel file(s) could not be deleted", e);
+    try {
+      if (isFile) {
+        deleteFile(bucket, key, bytesDeletedCalculator);
+      } else {
+        deleteFilesInDirectory(bucket, key, bytesDeletedCalculator);
+        deleteSentinelFiles(s3SchemeURI, key, bucket, housekeepingPath.getTableName());
+      }
+    } catch (Exception e) {
```
massdosage: Is this catch block necessary? Isn't the default behaviour to do what it's doing?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3PathCleaner.java
```diff
@@ -91,39 +86,57 @@ private S3SchemeURI extractURI(String housekeepingPath) {
     return s3SchemeURI;
   }
 
-  private void deleteFilesInDirectory(String bucket, String key) {
+  private void deleteFile(String bucket, String key, S3BytesDeletedCalculator bytesDeletedCalculator) {
+    bytesDeletedCalculator.cacheFiles(bucket, List.of(key));
```
massdosage: Looking at how it's being used, perhaps `storeFileSizes` is a better method name as it's not really a cache, the reason we're doing this is to store the file sizes before we delete them as after the delete that information is obviously lost.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3PathCleaner.java
```diff
@@ -91,39 +86,57 @@ private S3SchemeURI extractURI(String housekeepingPath) {
...
-    int totalFiles = objectsWithKeyAsPrefix.size();
-    int successfulDeletes = deleteAllObjectsAtKey(bucket, objectsWithKeyAsPrefix);
-
-    if (totalFiles != successfulDeletes) {
+    List<String> keys = s3Client.listObjects(bucket, key)
+      .stream()
+      .map(S3ObjectSummary::getKey)
+      .collect(Collectors.toList());
+    bytesDeletedCalculator.cacheFiles(bucket, keys);
+    List<String> keysDeleted = deleteFiles(bucket, keys);
```
massdosage: `keysDeleted` -> `deletedKeys`
(reads better IMO)

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/config/FileSystem.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.expediagroup.beekeeper.core.config;
+
+public enum FileSystem {
```
massdosage: `FileSystemType` ? It's not really a fully fledged `FileSystem`, this just describes the name/type of the FS?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporter.java
max-jacobs: Yes it could potentially be called by multiple threads. It looked like the `MeterRegistry` is thread safe.

Following the `register()` method through, we get to here: https://github.com/micrometer-metrics/micrometer/blob/f1fb973d6314cffac25059ee41926b487ba7bb25/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L548

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3PathCleaner.java
max-jacobs: Yes true, will remove the catch block

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporter.java
max-jacobs: So probably could even get away with not even searching for the counter, and just registering it each time?

massdosage: I don't know, wouldn't that register a new object on each report and thus you'd lose any previously stored values? (unless I'm misunderstanding you).

I think in our usage we do the deletes serially so I don't think we have multiple threads calling `report` so we're probably OK? Otherwise we'd need to make this synchronized or introduce locking.

max-jacobs: Indeed we do them serially. 

So in that link, the `getOrCreateMeter()` method uses synchronized to make sure that users can't create a new Meter while one is already being created. So if one thread finishes creating the same meter, then the next thread will just retrieve it. Does that make sense?

max-jacobs: And that ID used to identify a meter comprises the name, tags and meter type. So we're good on that front too as different each table will tag the meter differently.

massdosage: OK, yes, makes sense, so it sounds like we could call register each time or just leave the code like you have it.

max-jacobs: Haven't seen any examples of this no any documentation, so I think its probably fine to remove the check and just do the register.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporter.java
```diff
@@ -0,0 +1,60 @@
...
+  private String metricName;
+
+  @Autowired
+  public BytesDeletedReporter(MeterRegistry meterRegistry,
+    @Value("${properties.dry-run-enabled}") boolean dryRunEnabled) {
+    this.meterRegistry = meterRegistry;
+    this.metricName = dryRunEnabled ? DRY_RUN_METRIC_NAME : METRIC_NAME;
+  }
+
+  public void report(long bytesDeleted, String fullyQualifiedTableName, FileSystemType fileSystemType) {
```
nvitucci: If this metric is going to be used not just at table level, I would suggest to update it in one of the following ways:
- change the class name to something more specific, e.g. `TableLevelBytesDeletedReporter`;
- create more specific `report` methods, such as `reportTableLevel` etc.
- overload the `report` method and use different classes instead of a `String` name;
- update the `report` method to use a superclass or an interface instead of the `String` name.

max-jacobs: The last bullet point sounds good

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporter.java
```diff
@@ -0,0 +1,60 @@
...
+  public static final String DRY_RUN_METRIC_NAME = "dry-run-" + METRIC_NAME;
+
+  private MeterRegistry meterRegistry;
+  private String metricName;
+
+  @Autowired
+  public BytesDeletedReporter(MeterRegistry meterRegistry,
+    @Value("${properties.dry-run-enabled}") boolean dryRunEnabled) {
+    this.meterRegistry = meterRegistry;
+    this.metricName = dryRunEnabled ? DRY_RUN_METRIC_NAME : METRIC_NAME;
```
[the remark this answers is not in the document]
max-jacobs: I can't see this being used to count anything else at the moment. We can always refactor down the line.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporter.java
max-jacobs: @abhimanyugupta07 here's the discussion about creating the counter

PR #28, opened 2019-12-16
Event whitelisting

max-jacobs:
Beekeeper should only action `alter_partition` and `alter_table` events. Everything else needs to be whitelisted.


README.md
```diff
@@ -29,6 +29,16 @@ Beekeeper comprises two separate Spring-based Java applications. One application
...
 
+## Supported events
+
+Beekeeper will currently only be triggered by these events:
+- `alter_partition`
+- `alter_table`
+- `drop_partition`
+- `drop_table`
+
+By default, `alter_partition` and `alter_table` events require no further configuration. However, in order to avoid unexpected data loss, other even types require whitelsiting on a per table basis. See [Hive table configuration](#hive-table-configuration) for more details.
```
JayGreeeen: even -> event

massdosage: whitelsiting -> whitelisting

README.md
```diff
@@ -37,6 +47,7 @@ Beekeeper only actions on events which are marked with a specific parameter. Thi
 |:----|:----:|:----:|:----|
 | `beekeeper.remove.unreferenced.data=true`   | Yes |  `true` or `false`       | Set this parameter to ensure Beekeeper monitors your table for orphaned data. |
 | `beekeeper.unreferenced.data.retention.period=X` | No | e.g. `P7D` or `PT3H` (based on [ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601)) | Set this parameter to control the delay between schedule and deletion by Beekeeper. If this is either not set, or configured incorrectly, the default will be used. Default is 3 days. |
+| `beekeeper.hive.event.whitelist=X` | No | Comma separate list of event types to whitelist. For example, `beekeeper.hive.event.whitelist=drop_table,drop_partition`.  | Beekeeper will always process `alter_partition` and `alter_table` events. Add other event types to this list to ensure Beekeeper does not ignore them. |
```
massdosage:
```suggestion
| `beekeeper.hive.event.whitelist=X` | No | Comma separated list of event types to whitelist. For example, `beekeeper.hive.event.whitelist=drop_table,drop_partition`.  | Beekeeper will always process `alter_partition` and `alter_table` events. Add other event types to this list to ensure Beekeeper does not ignore them. |
```

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/MetadataOnlyListenerEventFilter.java
```diff
@@ -27,6 +27,9 @@
 
   @Override
   public boolean filter(ListenerEvent listenerEvent) {
+    if (listenerEvent == null) {
```
massdosage: When does this happen?

README.md
max-jacobs: Thanks both :)

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/MetadataOnlyListenerEventFilter.java
max-jacobs: This could happen if there's some problem with message desrialisation. It _probably_ wont ever happen.

We currently only have this check in `EventTypeListenerEventFilter`. All the filters are gathered in a list and executed one by one. This means that the `EventTypeListenerEventFilter` currently has to be first - and we've got no unit test for that. So I thought I'd add this check to all of the filters, and not have to rely on the ordering of them to catch edge cases.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [1.1.7] - TBD
 ### Adding
 - `TimedTaggable` annotation to time and report table level metrics.
+- `WhitelistedListenerEventFilter` to filter events unless listed in `beekeeper.hive.events.whitelist`.
```
massdosage: Let's go with 1.2.0 as the version number since this is a new feature, not a bugfix.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/MetadataOnlyListenerEventFilter.java
massdosage: Could we not do the null check in FilteringMessageReader and not call the Filters if its null? i.e. guarantee the filters that we'll never call them with null? I don't really mind and I see the value of having a check here but on the other hand I don't like that we now have to remember to put this null check into every single filter - too much repeated code.

max-jacobs:
@massdosage good plan, pushed this change

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/FilteringMessageReader.java
```diff
@@ -40,15 +41,26 @@ public FilteringMessageReader(MessageReader delegate, List<ListenerEventFilter>
...
   }
 
+  private boolean isFiltered(MessageEvent messageEvent) {
+    ListenerEvent listenerEvent = messageEvent.getEvent();
+    if (listenerEvent == null) {
+      return true;
+    }
+    boolean isFiltered = filters.stream()
+        .anyMatch(filter -> filter.filter(listenerEvent));
+    if (isFiltered) {
```
massdosage: You can just `return isFiltered` and remove this if statement.

README.md
```diff
@@ -37,6 +47,7 @@ Beekeeper only actions on events which are marked with a specific parameter. Thi
 |:----|:----:|:----:|:----|
 | `beekeeper.remove.unreferenced.data=true`   | Yes |  `true` or `false`       | Set this parameter to ensure Beekeeper monitors your table for orphaned data. |
 | `beekeeper.unreferenced.data.retention.period=X` | No | e.g. `P7D` or `PT3H` (based on [ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601)) | Set this parameter to control the delay between schedule and deletion by Beekeeper. If this is either not set, or configured incorrectly, the default will be used. Default is 3 days. |
+| `beekeeper.hive.event.whitelist=X` | No | Comma separated list of event types to whitelist. For example, `beekeeper.hive.event.whitelist=drop_table,drop_partition`.  | Beekeeper will always process `alter_partition` and `alter_table` events. Add other event types to this list to ensure Beekeeper does not ignore them. |
```
patduin: Not sure if this is the most flexible, perhaps this should just be overriding?
default value = `alter_partition,alter_table`
User wants to also drop_partition he needs to set:
`beekeeper.hive.event.whitelist=alter_partition,alter_table,drop_partition`
What would be more user friendly? Currently we can never ignore `alter_table` events if anyone would ever like that?
thoughts?

PR #30, opened 2019-12-19
[WIP] Lifecycle changes

KenFigueiredo:
PR to add Lifecycle changes (e.g. Table / Partition expirations) as per acceptance criteria:

> Add functionality to [Beekeeper] to allow a schema to be configured for lifecycle management (aka data retention).  This would allow certain schemas to automatically expire and have their Hive metastore metadata and S3 data automatically deleted, thus saving money for unwanted and unused data.

Currently a WIP - but here's a quick rundown of the work required & state:

* [x] Allow for path-scheduler-apiary to handle the ApiaryEvents[ADD_PARTITION, CREATE_TABLE]
* [x] Allow for expiration on ADD_PARTITION, ALTER_PARTITION, ALTER_TABLE, CREATE_TABLE events
* [ ] Properly cleanup expire events


beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/MessageEventToPathEventMapper.java
```diff
@@ -17,84 +17,141 @@
...
-import com.expediagroup.beekeeper.core.model.EntityHousekeepingPath;
-import com.expediagroup.beekeeper.core.model.PathStatus;
-import com.expediagroup.beekeeper.scheduler.apiary.model.PathEvent;
+import javax.annotation.PostConstruct;
+
+import static com.expediagroup.beekeeper.core.model.LifeCycleEventType.EXPIRED;
+import static com.expediagroup.beekeeper.core.model.LifeCycleEventType.UNREFERENCED;
 
 @Component
 public class MessageEventToPathEventMapper {
```
max-jacobs: It looks this would couple mapping/extracting the two event types together, and make it difficult to alter this. So I think probably this class could be split in to two, both which extend a `Mapper` interface -> `OrphanedEventMapper` + `ExpiredEventMapper` `implements MessageEventToPathMapper`
* `MessageReaderAdapter` can then loop through the mappers and add the housekeeping paths to the `BeekeeperEvent`, and at the end add the `MessageEvent`, so it actually becomes more of an adapter, and not just a proxy to call `map` on another class
* I think this change may make the `LifeCycleConfiguration` a little bit redundant, but would have to see what it looks like first.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/MessageEventToPathEventMapper.java
```diff
@@ -17,84 +17,141 @@
...
-      break;
-    default:
-      log.error("Unexpected event type: " + eventType);
-      break;
-    }
+            .pathStatus(PathStatus.SCHEDULED)
+            .creationTimestamp(LocalDateTime.now())
+            .cleanupDelay(extractCleanupDelay(listenerEvent, lifeCycleType))
+            .lifeCycleType(lifeCycleType.name())
+            .apiaryEventType(listenerEvent.getEventType().name())
```
max-jacobs: Why are we now saving the event type with the path?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/EventTypeTableListenerEventFilter.java
```diff
@@ -0,0 +1,61 @@
...
+import static com.expediagroup.beekeeper.core.model.LifeCycleEventType.EXPIRED;
+
+import java.util.Map;
+
+@Component
+public class EventTypeTableListenerEventFilter implements ListenerEventFilter {
+
+  @Override
+  public boolean filter(ListenerEvent listenerEvent) {
+    if (listenerEvent == null) {
```
max-jacobs: I think this null check is now done in the `FilteringMessageReader` so can be removed

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/context/CommonBeans.java
```diff
@@ -76,14 +85,43 @@ public S3BytesDeletedReporter s3BytesDeletedReporter(S3Client s3Client, MeterReg
...
+
+  @Bean
+  public HivePathCleaner hivePathCleaner(
+      HiveClient hiveClient
+  ) {
+    return new HivePathCleaner(hiveClient);
+  }
+
+  @Bean
+  public EnumMap<LifeCycleEventType, PathCleaner> pathCleanerMap(
```
max-jacobs: This `EnumMap` looks like a factory and is pretty much acting as one, so perhaps we should just have a factory bean instead?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3PathCleaner.java
```diff
@@ -77,8 +77,10 @@ public void cleanupPath(HousekeepingPath housekeepingPath) {
         }
       } catch (Exception e) {
         log.warn("Sentinel file(s) could not be deleted", e);
+        return false;
```
max-jacobs: I guess this case was left so that we don't fail the path due to a sentinel file error (which are zero bytes anyway). But perhaps it is right to. What do others think about this?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/hive/HivePathCleaner.java
```diff
@@ -0,0 +1,34 @@
...
+public class HivePathCleaner implements PathCleaner {
+    private static final Logger log = LoggerFactory.getLogger(HivePathCleaner.class);
+    private HiveClient hiveClient;
+
+    public HivePathCleaner(HiveClient hiveClient) {
+        this.hiveClient = hiveClient;
+    }
+
+    @Override
+    public boolean cleanupPath(HousekeepingPath housekeepingPath) {
```
max-jacobs: We can time this method too using `@TimedTaggable`

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/repository/HousekeepingPathRepository.java
```diff
@@ -34,4 +37,14 @@
...
   Page<EntityHousekeepingPath> findRecordsForCleanupByModifiedTimestamp(@Param("instant") LocalDateTime instant,
       Pageable pageable);
+
+  @Modifying
+  @Transactional
+  @Query("delete from EntityHousekeepingPath p where "
+      + " p.databaseName=:databaseName and "
+      + " p.tableName=:tableName and "
+      + " p.lifecycleType = 'EXPIRED'")
+  void cleanupOldExpiredRows(
```
max-jacobs: As per the other comment, this should be an update operation instead

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/LifeCycleEventType.java
```diff
@@ -0,0 +1,20 @@
+package com.expediagroup.beekeeper.core.model;
+
+import java.util.Map;
+
+public enum LifeCycleEventType {
```
max-jacobs: I like this enum, this encapsulates the beekeeper annotations nicely

beekeeper-path-scheduler-apiary/src/main/resources/beekeeper-path-scheduler-apiary-application.yml
```diff
@@ -11,5 +11,7 @@ spring:
 properties:
   beekeeper:
     default-cleanup-delay: P3D
+    default-expired-cleanup-delay: P356D
   apiary:
     cleanup-delay-property-key: beekeeper.unreferenced.data.retention.period
+    expired-cleanup-delay-property-key: beekeeper.expired.data.retention.period
```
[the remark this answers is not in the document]
max-jacobs: Actually, I'm not sure about this, what do others think about naming?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/EventTypeTableListenerEventFilter.java
```diff
@@ -0,0 +1,61 @@
...
+
+    Map<String, String> tableParameters = listenerEvent.getTableParameters();
+    if (tableParameters == null) {
+      return true;
+    }
+
+    Boolean isUnreferenced = UNREFERENCED.getBoolean(tableParameters);
+    Boolean isExpired = EXPIRED.getBoolean(tableParameters);
+
+    EventType eventType = listenerEvent.getEventType();
```
max-jacobs: Would be good to agree on what these scenarios are, in my mind `CREATE_TABLE` and `ALTER_TABLE` would insert a new expired row, and `ADD_PARTITION`, `ALTER_PARTITION` and `ALTER_TABLE` would update expired rows with a new deletion timestamp.

I think that `CREATE_TABLE` wouldn't necessarily have the required parameter we need on the table to check that we need to monitor it. We could either (1) require this parameter is added to the database and make sure Apiary emits these parameters or (2) maintain a list databases on Beekeeper which are marked to be monitored for expired data. (1) is definitely the nicer option but will need some extra work outside of Beekeeper.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/MessageEventToPathEventMapper.java
max-jacobs: As per your comment on line 90 in `MessageEventToPathEventMapper`, I think this will be tricky as we'll need to pass the `tableParameters` to the enum method to do this, but then we'll need to pass the keys too, which will make effort this redundant. Hence why this suggestion in the comment above potentially means splitting `ApiaryLifeCycleEvent` into two classes - one which returns an expired and one which returns an orphaned path.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/MessageEventToPathEventMapper.java
KenFigueiredo: The idea was for being able to have a distinction between a `CREATE_TABLE` vs `CREATE_PARTITION` expiration for cleanup since it seemed like those cases needed to be handled separately. 

My first thought on this was to make it a boolean flag (i.e. `isTable`), but was concerned if there were any future features we were to add that needed to drive different clean up types based on the apiary event, we could capture it w/ the event and not have to alter it later on.

I'm unsure if there is an existing/preferred way to make that distinction, but I can change this as needed.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/MessageEventToPathEventMapper.java
KenFigueiredo: 100% understand that reasoning. I've split that out into the suggested pattern w/ an abstract pathMapper and the Orphaned/Expired PathMappers extending it to share some utility methods between them.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/context/CommonBeans.java
KenFigueiredo: Replaced w/ an autowired list that populates all components of `MessageEventMapper`

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/BeekeeperPathSchedulerApiary.java
```diff
@@ -17,6 +17,7 @@
 
 import java.util.TimeZone;
 
+import com.expediagroup.beekeeper.scheduler.apiary.app.PathSchedulerApiaryRunner;
```
massdosage: This import needs to be back to where it was, we put our imports after third party imports.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/context/CommonBeans.java
```diff
@@ -17,6 +17,13 @@
 
 import java.util.List;
 
+import com.expediagroup.beekeeper.scheduler.apiary.filter.*;
```
massdosage: Imports need to be fully qualified and in the right order (we need to document this setup for OSS...)

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/EventTypeTableListenerEventFilter.java
massdosage: Good question. It depends a bit how we expect this all to be used. Are we going to say there will be a "scratch" database and *all* tables in this DB will have automatic expiry? I'm not sure which DB events we would listen to - there are create and drop events for DB's but I think that's about it. So we might have to go with option 2.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/FilteringMessageReader.java
```diff
@@ -53,8 +53,20 @@ private boolean isFiltered(MessageEvent messageEvent) {
     if (listenerEvent == null) {
       return true;
     }
-    return filters.stream()
-        .anyMatch(filter -> filter.filter(listenerEvent));
+
+    Boolean shouldFilter = false;
+
+    for(ListenerEventFilter f : filters) {
```
massdosage:
```suggestion
    for(ListenerEventFilter filter : filters) {
```

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/EventTypeTableListenerEventFilter.java
max-jacobs: It was my understanding that Hive can emit database parameters along with table events. If this is true, and we can get the Apiary listener sending this information along with create table events, then we could check for this database parameter without the need for maintaining a list on the service itself, or for monitoring create/drop database events.

massdosage: Ah right, yes, we'd need to try it out but that could work.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/MessageEventToPathEventMapper.java
max-jacobs: What is the distinction? As I understand it, `CREATE_TABLE` can create a new expired path entry, and `CREATE_PARTITION` can only update the expiration time on one that already exists.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/service/PathSchedulerApiary.java
```diff
@@ -18,39 +18,57 @@
...
 
 import com.expediagroup.beekeeper.core.error.BeekeeperException;
 import com.expediagroup.beekeeper.core.model.HousekeepingPath;
 import com.expediagroup.beekeeper.scheduler.apiary.messaging.PathEventReader;
-import com.expediagroup.beekeeper.scheduler.apiary.model.PathEvent;
+import com.expediagroup.beekeeper.scheduler.apiary.model.PathEvents;
 import com.expediagroup.beekeeper.scheduler.service.SchedulerService;
 
 @Component
 public class PathSchedulerApiary {
```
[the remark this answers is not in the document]
KenFigueiredo: That makes sense, having a better separation or responsibility for the orchestrator should make that easier to follow. Currently going through and trying to decouple that based on these comments, as well as adding the update rather than del/create

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/MessageEventToPathEventMapper.java
KenFigueiredo: My mistake - I misread the documentation. There shouldn't be a difference there, removing the addition

beekeeper-cleanup/pom.xml
```diff
@@ -86,6 +92,12 @@
       <version>${s3mock.version}</version>
       <scope>test</scope>
     </dependency>
+<!--    <dependency>-->
```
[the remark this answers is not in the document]
KenFigueiredo: Not at all - this is a net new dependency that I had commented it out as I was hitting some issues with integration tests due to java version incompatibility. 

This may be a good point to bring up a question I had on what the target Java version is for beekeeper. From what I could gather from the pom / docker image, beekeeper was targeted at Java 11, but the HMS library seems to be limited to Java 7/8. 

If that assumption / doc is correct, are there any alternatives to that library that could be used to accomplish the table deletes in hive? or would we need to refactor beekeeper to be compliant with Java 8?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/context/CommonBeans.java
```diff
@@ -76,14 +85,43 @@ public S3BytesDeletedReporter s3BytesDeletedReporter(S3Client s3Client, MeterReg
   }
 
   @Bean
-  public PathCleaner pathCleaner(S3Client s3Client, S3BytesDeletedReporter s3BytesDeletedReporter) {
+  public S3PathCleaner s3PathCleaner(S3Client s3Client, S3BytesDeletedReporter s3BytesDeletedReporter) {
     return new S3PathCleaner(s3Client, new S3SentinelFilesCleaner(s3Client), s3BytesDeletedReporter);
   }
 
+//  @Bean
```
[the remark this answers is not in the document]
KenFigueiredo: Yup! - This is still very much a WIP

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/context/CommonBeans.java
KenFigueiredo: I can update the contributing doc for this - is there a linter or editorconfig that we use standard to enforce a specific code style?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/context/CommonBeans.java
KenFigueiredo: 🤦‍♂  - misread this filename and thought this was in `beekeeper-path-scheduler-apiary`. The equivalent has been done there - will need to update this when I switch gears to the cleanup side

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/repository/HousekeepingPathRepository.java
KenFigueiredo: I replaced this with an incode upsert given that we'd most likely only want to update the
path, cleanupTimestamp or cleanupDelay fields. I had some sql written out for the upsert, but it seemed a bit convoluted and harder to maintain longterm than wrapping the functionality within a method.

It's at `PathSchedulerService#scheduleExpiration` now.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/EntityHousekeepingPath.java
```diff
@@ -267,6 +294,16 @@ public Builder cleanupAttempts(int cleanupAttempts) {
       return this;
     }
 
+    public Builder lifeCycleType(String cleanupType) {
```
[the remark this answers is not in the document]
KenFigueiredo: updated all occurrences to `lifecycleType`

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/FilteringMessageReader.java
KenFigueiredo: This was some leftover debug logic - removed & reverted to what was there previously

KenFigueiredo: This was some leftover debug logic - removed & reverted to what was there previously

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/context/CommonBeans.java
massdosage: We didn't have anything to do import ordering but after reviewing this I remembered seeing it in another project of ours so I have raised a PR against the parent pom (which this project uses) to sort that out - https://github.com/ExpediaGroup/eg-oss-parent/pull/8. Once that's merged we can upgrade the parent pom here and then all the import ordering and optimisation should be sorted out. We have a very high level editforconfig file at https://github.com/ExpediaGroup/new-project/blob/master/.editorconfig which we could add to this project if it would be useful although we've found it pretty limited in what it supports.

beekeeper-cleanup/pom.xml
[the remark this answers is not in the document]
massdosage: Good question. So... we wrote Beekeeper using Java 11 as initially it didn't have any Hive dependencies. With a related project we ran into the issue you describe with some code in Hive that isn't compatible with Java >=9. We raised a PR against Hive which has been merged into the necessary branches, we're just waiting for a Hive 2.3.7 release with those changes in it. We have an internally patched version that we've been using up to now but that's obviously not ideal and will be problematic for a production release. We should reach out to the Hive community and see if we can convince them to do a release which we can then use.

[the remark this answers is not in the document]
KenFigueiredo: Ok awesome - glad that we have a way forward with it. If we aren't able to get a release for it anytime soon, I can use the internal version in the meantime to get the feature up and running. Is there a specific route that we usually go in reaching out to the Hive maintainers? or is it just through their Jira?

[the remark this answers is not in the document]
massdosage: I commented on a recent email thread asking about a 2.4.0 release (see http://mail-archives.apache.org/mod_mbox/hive-user/201912.mbox/%3CCAE-i9zvR9s26DAz14Ywzuo99vUcGq_gtBHO83TuMV0vaw_8LyQ%40mail.gmail.com%3E) but I don't know if that will go anywhere. If you want to try help here please add a comment on https://issues.apache.org/jira/browse/HIVE-21508 asking when this will go into a Hive 2.3 release and then also put a post on the user email list referencing the Jira and asking about a release etc. The more people that ask the more likely it is they will do it.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/hive/HivePathCleaner.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    @TimedTaggable("hive-paths-deleted")
+    public boolean cleanupPath(HousekeepingPath housekeepingPath) {
+        if (LocalDateTime.now().isBefore(housekeepingPath.getCleanupTimestamp())) {
+            log.debug("Not deleting path \"{}\", expiration timestamp not met yet.", housekeepingPath.getPath());
+            return false;
+        }
+
+        // TODO: Determine difference between partition / table from houseKeepingPath and drop accordingly
```
max-jacobs: Not sure if this TODO is stale, but we're only dropping expired tables right?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/context/CommonBeans.java
```diff
@@ -47,6 +44,9 @@
 @EnableRetry(proxyTargetClass = true)
 public class CommonBeans {
 
+  @Autowired private List<MessageEventMapper> eventMappers;
```
max-jacobs: That's really nice, much cleaner

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/hive/HivePathCleaner.java
KenFigueiredo: Yeah this is stale and needs to be cleaned up still

KenFigueiredo:
Closing this out in favor of a dialed back PR including just the refactor changes

PR #31, opened 2020-01-03
Fix s3 space encoding

max-jacobs:
This PR replaces findify with localstack (the same library we're already using in integration tests) as I was seeing some inconsistencies with how findify deals with paths with spaces. Localstack's S3 mock appears to handle spaces more true to AWS.

The main code change is in the `S3SchemeUri` and `S3SentinelFilesCleaner` - both now encode the urls so spaces shouldn't be a problem. `S3SchemeUri` also uses regex instead of java in built `URI` class to parse `s3a` and `s3n` paths.

I've also added a bunch of tests.


beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
```diff
@@ -17,29 +17,27 @@
...
 
-  public S3SchemeURI(String housekeepingPath) throws URISyntaxException {
+  public S3SchemeURI(String housekeepingPath) {
     if (!housekeepingPath.startsWith("s3")) {
       throw new BeekeeperException(format("'%s' is not an S3 path.", housekeepingPath));
     }
-    URI uri = URI.create(housekeepingPath);
-    URI s3Uri = new URI("s3", uri.getHost(), uri.getPath(), null);
-    this.amazonS3URI = new AmazonS3URI(s3Uri);
+    String s3Path = housekeepingPath.replaceFirst(S3_SCHEME_REGEX, S3_SCHEME);
```
nvitucci: So a scheme replacement is all that is needed? Have you found any docs/messages stating this?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SentinelFilesCleaner.java
```diff
@@ -32,7 +32,7 @@ public S3SentinelFilesCleaner(S3Client s3Client) {
 
   @Override
   public void deleteSentinelFiles(String absolutePath) {
-    AmazonS3URI s3Path = new AmazonS3URI(absolutePath, false);
+    AmazonS3URI s3Path = new AmazonS3URI(absolutePath, true);
```
nvitucci: Is this flipped to `true` to let the encoder deal with spaces and other odd formats?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperCleanupIntegrationTest.java
```diff
@@ -167,6 +167,26 @@ void cleanupPathsForDirectory() throws SQLException {
     assertThat(amazonS3.doesObjectExist(BUCKET, OBJECT_KEY_OTHER_SENTINEL)).isTrue();
   }
 
+  @Test
+  void cleanupPathsForDirectoryWithSpace() throws SQLException {
+    String objectKeyRoot = "database/table/ /id1/partition1";
+    String objectKey1 = objectKeyRoot+ "/file1";
```
nvitucci: Space between `objectKeyRoot` and `+`

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SentinelFilesCleaner.java
max-jacobs: Exactly

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
max-jacobs: Good question, I'm pretty sure they are interchangeable, but a quick look for docs doesn't reveal much.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
```diff
@@ -17,29 +17,27 @@
...
+    this.amazonS3URI = new AmazonS3URI(s3Path);
   }
 
   public String getPath() {
-    return amazonS3URI.getURI()
-      .toString();
+    return "s3://" + amazonS3URI.getBucket() + "/" + amazonS3URI.getKey();
```
massdosage: Just curious, what does `toString()` on `AmazonS3URI` return? I'm guessing something not useful for what we need here?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
```diff
@@ -17,29 +17,27 @@
...
-  public S3SchemeURI(String housekeepingPath) throws URISyntaxException {
+  public S3SchemeURI(String housekeepingPath) {
     if (!housekeepingPath.startsWith("s3")) {
       throw new BeekeeperException(format("'%s' is not an S3 path.", housekeepingPath));
     }
-    URI uri = URI.create(housekeepingPath);
-    URI s3Uri = new URI("s3", uri.getHost(), uri.getPath(), null);
-    this.amazonS3URI = new AmazonS3URI(s3Uri);
+    String s3Path = housekeepingPath.replaceFirst(S3_SCHEME_REGEX, S3_SCHEME);
+    this.amazonS3URI = new AmazonS3URI(s3Path);
```
massdosage: And we don't need to pass in the boolean `true` as an additional param like we do in the class below?

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/path/aws/AmazonS3Factory.java
```diff
@@ -15,27 +15,24 @@
  */
 package com.expediagroup.beekeeper.cleanup.path.aws;
 
-import akka.http.scaladsl.Http;
```
massdosage: Bye bye weird Scala classes!

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
max-jacobs: `toString()` returns the encoded path (i.e. with %20 etc), which I don't think is useful for us. We only want to deal with the unencoded path. I did some testing and the encoded path doesn't work with S3.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
max-jacobs: Nope, this is set automatically in this constructor

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/path/aws/S3ClientTest.java
```diff
@@ -16,57 +16,84 @@
...
 
-  private final S3Mock s3Mock = new S3Mock.Builder().withPort(0).withInMemoryBackend().build();
-  private AmazonS3 amazonS3;
   private S3Client s3Client;
   private S3Client s3ClientDryRun;
+  private AmazonS3 amazonS3;
+
+  @BeforeAll
+  public static void s3() {
+    s3Container = new LocalStackContainer().withServices(LocalStackContainer.Service.S3);
```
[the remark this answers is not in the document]
max-jacobs: Weird. Perhaps because eclipse can't see the Junit lifecycle?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
nvitucci: By looking [here|https://cwiki.apache.org/confluence/display/HADOOP2/AmazonS3] and [here|https://www.edureka.co/community/920/difference-between-s3n-s3a-and-s3], my understanding is that `s3a` is going to replace `s3n` and simply abstracts `s3` (which is block-based) to use it as an object-based filesystem. I can't understand why the switch would be necessary though.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/path/aws/S3ClientTest.java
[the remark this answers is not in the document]
abhimanyugupta07: Perhaps. I wasn't able to find much on internet.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
abhimanyugupta07: In a recent AWS support case, I was told that we should only be using `s3://`.

https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-plan-file-systems.html

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3SchemeURI.java
massdosage: OK

PR #35, opened 2020-01-08
Batch delete

max-jacobs:
PR to do batch deletes of partition files, with a little refactoring in as I found that `S3ObjectSummary` contains the file size - so no need to re-lookup.

### DIFF
--- a/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedCalculator.java
+++ b/beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedCalculator.java
@@ -21,8 +21,6 @@ import java.util.Map;
 
 import com.amazonaws.services.s3.model.S3ObjectSummary;
 
-import com.expediagroup.beekeeper.core.error.BeekeeperException;
-
 public class S3BytesDeletedCalculator {
 
   private S3Client s3Client;
@@ -34,13 +32,11 @@ public class S3BytesDeletedCalculator {
   }
 
   public void storeFileSize(String bucket, String key) {
-    checkKeysAreEmpty();
     long bytes = s3Client.getObjectMetadata(bucket, key).getContentLength();
     keyToSize.put(key, bytes);
   }
 
   public void storeFileSizes(List<S3ObjectSummary> objectSummaries) {
-    checkKeysAreEmpty();
     objectSummaries.forEach(objectSummary -> {
       long bytes = objectSummary.getSize();
       keyToSize.put(objectSummary.getKey(), bytes);
@@ -59,12 +55,6 @@ public class S3BytesDeletedCalculator {
 
   public long getBytesDeleted() {
     return bytesDeleted;
-  }
-
-  private void checkKeysAreEmpty() {
-    if (!keyToSize.isEmpty()) {
-      throw new BeekeeperException("Should not store file sizes twice.");
-    }
   }
 
 }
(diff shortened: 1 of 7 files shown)


beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3BytesDeletedCalculator.java
```diff
@@ -55,4 +61,10 @@ public long getBytesDeleted() {
     return bytesDeleted;
   }
 
+  private void checkKeysAreEmpty() {
+    if (!keyToSize.isEmpty()) {
+      throw new BeekeeperException("Should not store file sizes twice.");
```
nvitucci: I don't see this exception caught anywhere, so is this an error that would make the whole process to stop?

patduin: +1 this looks odd when does this happen, perhaps we should just log a warning instead of throwing an exception?

max-jacobs: Yep happy to log instead. Just wanted to enforce this class only being used once before being thrown away as in my head there's no "add keys" option, but actually that's exactly what the method does. So it's actually fine if this method is reused.

max-jacobs: I've removed the exception, I didn't put the log in either, seemed unnecessary. Happy to put it in though if you prefer

patduin: looks ok

patduin:
can you add a changlog, also this doesn't actually has much to do with batch deletions right it's just  calculating the file size in batch correct?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/path/aws/S3Client.java
```diff
@@ -50,11 +52,20 @@ void deleteObject(String bucket, String key) {
   }
 
   List<S3ObjectSummary> listObjects(String bucket, String key) {
```
max-jacobs: @patduin It does have to do with batch deletions here. `S3Client.listObjects(...)` now paginates through the results to return everything at a key.

max-jacobs: A possible issue with this implementation is that if we have a drop table and there are 10's or 100s of thousands of files, this could fail

patduin: right I see I got distracted by the byte calculating stuff. Why would it fail?

max-jacobs: Possibly because the batching is within the `listObjects` method only, so this gets all of the objects first and processes them all second. Instead of getting a batch, processing them, and then getting the next batch and processing them. Potentially could be too heavy to handle it all at once.

patduin: ok so you want to go with this approach first and if it fails we'll deal with it?

max-jacobs: Yep, this is the plan

PR #36, opened 2020-01-12
Refactor Beekeeper to support future lifecycle changes

KenFigueiredo:
This PR captures the refactor specific portions from https://github.com/ExpediaGroup/beekeeper/pull/30 to prep the codebase to support lifecycle changes (aka expired) in the future once we have better Java 11 HMS support. 

### Added
- Add `LifecycleEventType` enum in `beekeeper-core` to describe supported data lifecycles.
- Increase unit test coverage for `beekeeper-path-scheduler-apiary` package.
- Increase coverage for integration tests.

### Changed
- Refactored internals of `beekeeper-path-scheduler-apiary` to support generic data lifecycle scheduling.
    - Inserted additional workflow (handlers) between read & filter actions to support filters per lifecycle type.
    - MessageReaderAdapter now has additional logic to orchestrate the updated workflow.
    - Rename `PathEvent` to `BeekeeperEvent` to better reflect event types.
- Refactored internals of `beekeeper-path-scheduler` to support generic data lifecycle scheduling.
    - Rename `PathSchedulerService` to `UnreferencedPathSchedulerService` to differentiate types
- Refactored internals of `beekeeper-cleanup` to support generic data lifecycle data lifecycle deletions.
    - Refactor `PagingCleanupService` to be a generic orchestrator of Lifecycle handlers.


beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
```diff
@@ -0,0 +1,160 @@
...
+    return generateHouseKeepingPaths(listenerEvent);
+  }
+
+  /**
+   * Get's a boolean flag if the given tableparameters exist for this mapper
+   *
+   * @param tableParameters
+   * @return boolean
+   */
+  boolean checkIfTablePropertyExists(Map<String, String> tableParameters) {
```
max-jacobs: It looks like this method is duplicating what the `TableParameterListenerEventFilter` class does?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
```diff
@@ -0,0 +1,160 @@
...
+import com.expediagroup.beekeeper.core.model.PathStatus;
+import com.expediagroup.beekeeper.scheduler.apiary.filter.FilterType;
+import com.expediagroup.beekeeper.scheduler.apiary.filter.ListenerEventFilter;
+import com.expediagroup.beekeeper.scheduler.apiary.model.EventModel;
+
+@Component
+public abstract class MessageEventHandler {
+
+  private static final Logger log = LoggerFactory.getLogger(MessageEventHandler.class);
+  private final String CLIENT_ID = "apiary-metastore-event";
```
max-jacobs: This can be static too

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/TableParameterListenerEventFilter.java
```diff
@@ -25,6 +25,10 @@
 public class TableParameterListenerEventFilter implements ListenerEventFilter {
 
   private static final String BEEKEEPER_TABLE_PARAMETER = "beekeeper.remove.unreferenced.data";
```
max-jacobs: This could be replaced by the lifecycle event type enum

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/MessageReaderAdapter.java
```diff
@@ -16,38 +16,58 @@
...
+    }
+
+    MessageEvent message = messageEvent.get();
+
+    List<HousekeepingPath> housekeepingPaths = handlers.parallelStream()
+        .map(eventHandler -> eventHandler.handleMessage(message))
+        .flatMap(x -> x.stream())
+        .collect(Collectors.toList());
+
+    if (housekeepingPaths.size() <= 0) {
```
max-jacobs: If there are no housekeeping paths, we can return an empty optional, but we still need to remove the event from the queue

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/messaging/MessageReaderAdapter.java
```diff
@@ -16,38 +16,58 @@
...
+
+    if (!messageEvent.isPresent()) {
+      return Optional.empty();
+    }
+
+    MessageEvent message = messageEvent.get();
+
+    List<HousekeepingPath> housekeepingPaths = handlers.parallelStream()
+        .map(eventHandler -> eventHandler.handleMessage(message))
+        .flatMap(x -> x.stream())
```
max-jacobs: x could be paths?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/model/EventModel.java
```diff
@@ -0,0 +1,15 @@
+package com.expediagroup.beekeeper.scheduler.apiary.model;
+
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+
+public class EventModel {
+
+  public final LifecycleEventType lifecycleEvent;
```
max-jacobs: These could be private and have getters/setters

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/service/PathSchedulerApiary.java
```diff
@@ -18,45 +18,64 @@
...
-  public void schedulePath() {
-    Optional<PathEvent> pathToBeScheduled = pathEventReader.read();
-    if (pathToBeScheduled.isPresent()) {
-      PathEvent pathEvent = pathToBeScheduled.get();
-      HousekeepingPath path = pathEvent.getHousekeepingPath();
+  @Transactional
+  public void scheduleBeekeeperEvent() {
+    Optional<BeekeeperEvent> pathToBeScheduled = beekeeperEventReader.read();
+
+    if (!pathToBeScheduled.isPresent()) {
```
max-jacobs: Can remove the empty lines here. Also !isPresent() can be isEmpty()

beekeeper-path-scheduler-apiary/src/main/resources/db/migration/V1_2__Alter_path_add_lifecycle_type.sql
```diff
@@ -0,0 +1,2 @@
+USE beekeeper;
+ALTER TABLE path ADD COLUMN lifecycle_type VARCHAR(50) NOT NULL DEFAULT "UNREFERENCED";
```
max-jacobs: Need a new line here too

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/PathSchedulerApiaryIntegrationTest.java
```diff
@@ -0,0 +1,328 @@
...
+import com.expediagroup.beekeeper.core.model.PathStatus;
+import com.expediagroup.beekeeper.integration.model.AddPartitionSqsMessage;
+import com.expediagroup.beekeeper.integration.model.AlterPartitionSqsMessage;
+import com.expediagroup.beekeeper.integration.model.AlterTableSqsMessage;
+import com.expediagroup.beekeeper.integration.model.CreateTableSqsMessage;
+import com.expediagroup.beekeeper.integration.model.DropPartitionSqsMessage;
+import com.expediagroup.beekeeper.integration.model.DropTableSqsMessage;
+import com.expediagroup.beekeeper.scheduler.apiary.BeekeeperPathSchedulerApiary;
+
+public class PathSchedulerApiaryIntegrationTest {
```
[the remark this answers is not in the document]
KenFigueiredo: Not entirely sure to be honest - I may have unintentionally refactored the name w/ a bulk IDE change. Reverting this

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/ListenerEventFilter.java
```diff
@@ -19,6 +19,7 @@
 
 public interface ListenerEventFilter {
 
-  boolean filter(ListenerEvent listenerEvent);
+  FilterType getFilterType();
```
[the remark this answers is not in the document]
KenFigueiredo: It's used to setup a filterMap (FilterType -> Filter relation) that `MessageEventHandler` uses to get all validFilters for a given lifecycle event handler.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
```diff
@@ -0,0 +1,160 @@
...
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+import com.expediagroup.beekeeper.core.model.PathStatus;
+import com.expediagroup.beekeeper.scheduler.apiary.filter.FilterType;
+import com.expediagroup.beekeeper.scheduler.apiary.filter.ListenerEventFilter;
+import com.expediagroup.beekeeper.scheduler.apiary.model.EventModel;
+
+@Component
+public abstract class MessageEventHandler {
+
+  private static final Logger log = LoggerFactory.getLogger(MessageEventHandler.class);
```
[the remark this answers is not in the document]
KenFigueiredo: its used on line 153 of this to log if a cleanUpDelay cannot be extracted properly from a table parameter

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/MySqlTestUtils.java
```diff
@@ -52,43 +55,54 @@
...
 
-  public MySqlTestUtils(String jdbcUrl, String username, String password) throws SQLException {
+  MySqlTestUtils(String jdbcUrl, String username, String password) throws SQLException {
     connection = DriverManager.getConnection(jdbcUrl, username, password);
   }
 
-  public void insertPath(String path, String table) throws SQLException {
-    String fields = String.join(", ", PATH, PATH_STATUS, CLEANUP_DELAY, CLEANUP_TIMESTAMP, TABLE_NAME);
+  void insertPath(String path, String table) throws SQLException {
+    // TODO: for now, we'll just keep testing unreferenced paths only
```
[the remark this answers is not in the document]
KenFigueiredo: Nope! Just a left over artifact from the cherrypick. Removed!

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupService.java
```diff
@@ -20,34 +20,31 @@
...
-  private final PathCleaner pathCleaner;
+  private final List<GenericHandler> pathHandlers;
   private final boolean dryRunEnabled;
   private final int pageSize;
 
-  public PagingCleanupService(HousekeepingPathRepository housekeepingPathRepository, PathCleaner pathCleaner,
-      int pageSize, boolean dryRunEnabled) {
-    this.housekeepingPathRepository = housekeepingPathRepository;
-    this.pathCleaner = pathCleaner;
+  public PagingCleanupService(List<GenericHandler> pathHandlers, int pageSize, boolean dryRunEnabled) {
```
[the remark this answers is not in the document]
KenFigueiredo: removed pageHandler & updated

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
KenFigueiredo: yeah - good catch. This is another remnant of the previous PR. Removed.

It also brings up a design question: Should filters have context into what lifecycle event type they are filtering?

max-jacobs: For now at least, I think not. Whatever filters are injected in to each path handler are the ones that will be used, the reverse relationship doesn't need to be known and I can't think of a usecase for this in the future either. Do you have something in mind?

KenFigueiredo: Yeah, specifically for `TableParameterListenerEventFilter` the table param it checks is directly related to the lifecycle type from what I can tell. Given that, if we're using the enum for it, we'd have to pass that in directly from the filter call which is basically how I refactored it again from these PR comments.

.gitignore
```diff
@@ -32,3 +32,6 @@ target/
 
 # AWS credentials
 *.pem
+
+# HMS data
+metastore_db
```
max-jacobs: This needs a new line

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
+- Increase unit test coverage for `beekeeper-path-scheduler-apiary` package.
+- Increase coverage for integration tests.
+
+### Changed
+- Refactored internals of `beekeeper-path-scheduler-apiary` to support generic data lifecycle scheduling.
+    - Inserted additional workflow (handlers) between read & filter actions to support filters per lifecycle type.
+    - MessageReaderAdapter now has additional logic to orchestrate the updated workflow.
+    - Rename `PathEvent` to `BeekeeperEvent` to better reflect event types.
+- Refactored internals of `beekeeper-path-scheduler` to support generic data lifecycle scheduling.
+    - Rename `PathSchedulerService` to `UnreferencedPathSchedulerService` to differentiate types
```
max-jacobs:
```suggestion
    - Rename `PathSchedulerService` to `UnreferencedPathSchedulerService` to differentiate types.
```

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
+- Refactored internals of `beekeeper-cleanup` to support generic data lifecycle data lifecycle deletions.
```
max-jacobs:
```suggestion
- Refactored internals of `beekeeper-cleanup` to support generic data Lifecycle deletions.
```

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
+- Add `LifecycleEventType` enum in `beekeeper-core` to describe supported data lifecycles.
+- Increase unit test coverage for `beekeeper-path-scheduler-apiary` package.
+- Increase coverage for integration tests.
+
+### Changed
+- Refactored internals of `beekeeper-path-scheduler-apiary` to support generic data lifecycle scheduling.
+    - Inserted additional workflow (handlers) between read & filter actions to support filters per lifecycle type.
+    - MessageReaderAdapter now has additional logic to orchestrate the updated workflow.
+    - Rename `PathEvent` to `BeekeeperEvent` to better reflect event types.
+- Refactored internals of `beekeeper-path-scheduler` to support generic data lifecycle scheduling.
```
max-jacobs:
```suggestion
- Refactored internals of `beekeeper-path-scheduler` to support generic Lifecycle scheduling.
```

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
 
+## [2.0.0] - TBD
+### Added
+- Add `LifecycleEventType` enum in `beekeeper-core` to describe supported data lifecycles.
+- Increase unit test coverage for `beekeeper-path-scheduler-apiary` package.
+- Increase coverage for integration tests.
+
+### Changed
+- Refactored internals of `beekeeper-path-scheduler-apiary` to support generic data lifecycle scheduling.
+    - Inserted additional workflow (handlers) between read & filter actions to support filters per lifecycle type.
```
max-jacobs:
```suggestion
    - Inserted additional workflow (handlers) between read & filter actions to support filters per Lifecycle type.
```

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.0.0] - TBD
+### Added
+- Add `LifecycleEventType` enum in `beekeeper-core` to describe supported data lifecycles.
+- Increase unit test coverage for `beekeeper-path-scheduler-apiary` package.
+- Increase coverage for integration tests.
+
+### Changed
+- Refactored internals of `beekeeper-path-scheduler-apiary` to support generic data lifecycle scheduling.
```
max-jacobs:
```suggestion
- Refactored internals of `beekeeper-path-scheduler-apiary` to support generic Lifecycle scheduling.
```

CHANGELOG.md
max-jacobs:
```suggestion
    - Inserted additional workflow (handlers) between read & filter actions to support filters per lifecycle type.
``L

File: beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/handler/UnreferencedHandler.java [hunk 20]
```diff
@@ -0,0 +1,46 @@
...
+
+@Component
+public class UnreferencedHandler extends GenericHandler {
+
+  private static final LifecycleEventType EVENT_TYPE = LifecycleEventType.UNREFERENCED;
+  private S3PathCleaner s3PathCleaner;
+
+  @Autowired
+  public UnreferencedHandler(S3PathCleaner s3PathCleaner) {
+    super();
```
massdosage: I'm pretty sure the compiler will insert this for you so you should be able to remove it.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/handler/UnreferencedHandler.java
```diff
@@ -0,0 +1,46 @@
...
+    this.s3PathCleaner = s3PathCleaner;
+  }
+
+  @VisibleForTesting
+  public void setS3PathCleaner(S3PathCleaner cleaner) {
```
massdosage: Do you really need it for testing if you can pass this in to the constructor? If you do need it could it be made `protected` or default? instead of `public`?

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/handler/GenericHandlerTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Mock private HousekeepingPathRepository housekeepingPathRepository;
+  @Mock private S3PathCleaner pathCleaner;
+  @InjectMocks private final UnreferencedHandler handler = new UnreferencedHandler(pathCleaner);
+  @Mock private EntityHousekeepingPath mockPath;
+
+  @Before
+  public void init() {
+    handler.setS3PathCleaner(pathCleaner);
+    MockitoAnnotations.initMocks(this);
```
massdosage: Do you need to do this if you're using the `ExtendWith` annotation?

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
```diff
@@ -1,225 +1,83 @@
-/**
```
massdosage: This shouldn't be removed. I'd also suggest running the full Maven lifecycle over all the code as this should add the license headers to all the new files for you.

beekeeper-integration-tests/src/test/resources/add_partition.json
```diff
@@ -0,0 +1,11 @@
+{
```
massdosage: Hmmm, I guess we're doing all this because Beekeeper currently directly uses the SNS messages. At some time in the future I think we're going to change this to use "DroneFly" downstream of the Kafka events and then implement Beekeeper to look like just a normal HiveMetastoreEventListener. @max-jacobs or others, any ideas how we can try minimise the amount of code that is in this PR that is coupled to SNS while still testing all the paths?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/BeekeeperPathSchedulerApiary.java
```diff
@@ -76,4 +71,18 @@ public static void stop() {
   public static MeterRegistry meterRegistry() {
     return meterRegistry;
   }
+
+  @VisibleForTesting
+  public static void resetStaticContext() {
```
massdosage: I'm surprised this works with nothing having to be closed...

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/BeekeeperPathSchedulerApiaryTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @BeforeEach()
+  public void resetContext() {
+    BeekeeperPathSchedulerApiary.resetStaticContext();
+  }
+
+  @Test
+  public void verifyApplicationContext() {
+    setupContextMock(true, true, true);
+    return;
```
massdosage: no need for this `return`

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/context/CommonBeansTest.java
```diff
@@ -0,0 +1,80 @@
...
+  @Mock private MessageReader messageReader;
+
+  @AfterAll
+  static void teardown() {
+    System.clearProperty(AWS_REGION_PROPERTY);
+    System.clearProperty(AWS_S3_ENDPOINT_PROPERTY);
+  }
+
+  @BeforeEach
+  void setUp() {
```
massdosage: Can you swap `setUp()` and `tearDown()` method order please, it just reads a bit easier.

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/context/CommonBeansTest.java
```diff
@@ -0,0 +1,80 @@
...
+  private static final String REGION = "us-west-2";
+  private static final String AWS_ENDPOINT = String.join(".", "s3", REGION, "amazonaws.com");
+  private static final String ENDPOINT = "endpoint";
+  private static final String BUCKET = "bucket";
+  private static final String KEY = "key";
+  private final CommonBeans commonBeans = new CommonBeans();
+  @Mock private MessageReader messageReader;
+
+  @AfterAll
+  static void teardown() {
```
massdosage:
```suggestion
  static void tearDown() {
```

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/handler/MessageEventHandlerTest.java
```diff
@@ -0,0 +1,90 @@
...
+  @InjectMocks private final UnreferencedMessageHandler msgHandler = new UnreferencedMessageHandler(
+      UNREF_HIVE_KEY, UNREF_DEFAULT);
+  @Spy private final EnumMap<FilterType, ListenerEventFilter> filterMap = new EnumMap<>(FilterType.class);
+  @Mock private MessageEvent messageEvent;
+  @Mock private AlterPartitionEvent listenerEvent;
+  @Mock private WhitelistedListenerEventFilter whiteListFilter;
+  @Mock private TableParameterListenerEventFilter tableFilter;
+
+  @Before
+  public void setup() throws Exception {
```
massdosage:
```suggestion
  public void setUp() throws Exception {
```

massdosage: I think you can ditch this whole method since you're using the mockito extension?

pom.xml
```diff
@@ -25,7 +26,8 @@
 
   <scm>
     <connection>scm:git:https://${GIT_USERNAME}:${GIT_PASSWORD}@github.com/ExpediaGroup/beekeeper.git</connection>
-    <developerConnection>scm:git:https://${GIT_USERNAME}:${GIT_PASSWORD}@github.com/ExpediaGroup/beekeeper.git</developerConnection>
+    <developerConnection>scm:git:https://${GIT_USERNAME}:${GIT_PASSWORD}@github.com/ExpediaGroup/beekeeper.git
+    </developerConnection>
```
massdosage: Can this go back onto one line?

beekeeper-integration-tests/src/test/resources/add_partition.json
max-jacobs: @massdosage my understanding of Drone Fly is that, as you say, it will act as a layer downstream of Kafka, but also that it will still land messages on the Apiary SNS topic - in which case we will be using SNS/SQS. How will Beekeeper consume messages from drone fly as a normal HiveMetastoreEventListener?

max-jacobs: Do you mean that Beekeeper will eventually be pointed directly at Kafka? In which case the additions in `com.expediagroup.beekeeper.integration.model` seem to make the integration tests more extendible and if we can use them as a template for Kafka messages/tests then we could keep them and refactor later.

massdosage: Well, DroneFly will probably expose events via the Hive `MetaStoreEventListener` interface - so an idea could be that Beekeeper runs as a listener on this level and deals with the Hive *Event objects directly and doesn't need to know anything about Kafka and there won't be any need for SQS/SNS. We could decide to still use SQS and SNS like you suggest though, just not sure it would be needed (we probably need to think a bit about how to queue up and distribute the work though).

max-jacobs: Right this makes sense. So the path scheduler no longer runs as a service, but as a listener on the metastore. In which case there will be more to refactor than just these additions. For the time being I think these additions are probably fine as they seem to make testing quite easy/extendible. I'm happy to keep them in the knowledge that they will most likely be removed in the future.

massdosage: OK, sounds good.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/handler/UnreferencedHandler.java
```diff
@@ -0,0 +1,46 @@
...
+import com.expediagroup.beekeeper.cleanup.path.PathCleaner;
+import com.expediagroup.beekeeper.cleanup.path.aws.S3PathCleaner;
+import com.expediagroup.beekeeper.core.model.EntityHousekeepingPath;
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+
+@Component
+public class UnreferencedHandler extends GenericHandler {
+
+  private static final LifecycleEventType EVENT_TYPE = LifecycleEventType.UNREFERENCED;
+  private S3PathCleaner s3PathCleaner;
```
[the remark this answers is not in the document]
KenFigueiredo: Shouldn't be an issue. A question for more context on this though and related to the `HdfsPathCleaner` comment above, do we see the UnreferencedHandler needing to ever cleanup HDFS items? Or is it only ever going to cleanup S3 items?

I was under the assumption that when we introduce the ExpiredHandler it would need the ability to cleanup HDFS & S3 Paths and have them both referenced as separate cleaners, but Unreferenced would not need to have this distinction.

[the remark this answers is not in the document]
massdosage: Good question. Up to now we've only needed to support S3 but we introduced the interface and S3-specific implementation based on other projects where we've needed to support different file systems. So, I think where possible we should try to be FileSystem-agnostic and allow the FileSystem in use to be chosen by configuration - or at the very least to hard code this for now (e.g. in CommonBeans) but with the idea that it would be fairly easy to introduce another FileSystem later when needed. Hopefully most other classes can just rely on a generic PathCleaner and not need to know whether S3/HDFS/GCS etc. is in use.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/handler/GenericHandlerTest.java
KenFigueiredo: Apparently not. At some point in working on this, injectMocks didn't seem to be populating correctly without this `initMocks` call, but that seems like it was probably due to some other error that I was attributing this to 😅. Removed.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/handler/UnreferencedHandler.java
KenFigueiredo: Dug into this a bit more - it looks like the underlying issue was my IDE overriding PathCleaner to final which was breaking mockito's setup for the pathcleaner on some tests. Moved this to be just a generic protected setter which fixes this issue.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/BeekeeperPathSchedulerApiary.java
KenFigueiredo: 😅 yeah, not super happy with how this is setup but it works. I think I was able to get away with it as on first setup of the tests for these, every static variable is already mocked so we're effectively just blowing away if the mock is set or not for those contexts.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/handler/UnreferencedHandler.java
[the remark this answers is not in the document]
KenFigueiredo: Reverted to PathCleaner

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
KenFigueiredo: not sure how that got removed - either way added all license headers

KenFigueiredo:
@massdosage @max-jacobs 
Aside from the travis failure I believe I've hit all the comments barring an open question about how to handle filters. Are there any last checks / fixes I can ensure to get this merged?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/handler/UnreferencedHandler.java
```diff
@@ -0,0 +1,56 @@
...
+  @Override
+  public LifecycleEventType getLifecycleType() {
+    return EVENT_TYPE;
+  }
+
+  @Override
+  public PathCleaner getPathCleaner() { return pathCleaner; }
+
+  @Override
+  protected void setPathCleaner(PathCleaner pathCleaner) { this.pathCleaner = pathCleaner; }
```
max-jacobs: What's the purpose of this setter? Do we need this to support further lifecycle changes?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupService.java
```diff
@@ -56,46 +49,23 @@ public PagingCleanupService(HousekeepingPathRepository housekeepingPathRepositor
...
-          pageable);
-
-      while (!page.getContent().isEmpty()) {
-        processPage(page.getContent());
-        if (dryRunEnabled) {
-          pageable = pageable.next();
-        }
-        page = housekeepingPathRepository.findRecordsForCleanupByModifiedTimestamp(instant, pageable);
-      }
+      pathHandlers.stream().forEach(handler -> pagingCleanup(handler, referenceTime));
```
max-jacobs: `pathHandlers` can directly call `forEach()` here

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
```diff
@@ -16,210 +16,83 @@
...
-@TestPropertySource(properties = {
-    "hibernate.data-source.driver-class-name=org.h2.Driver",
-    "hibernate.dialect=org.hibernate.dialect.H2Dialect",
-    "hibernate.hbm2ddl.auto=create",
-    "spring.datasource.url=jdbc:h2:mem:beekeeper;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE;MODE=MySQL" })
-@ContextConfiguration(classes = { TestApplication.class },
-    loader = AnnotationConfigContextLoader.class)
-@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
-class PagingCleanupServiceTest {
+public class PagingCleanupServiceTest {
```
max-jacobs: A lot of the tests have been removed here. Why is this? 

I'm not entirely sure the paging will work with these changes. I remember we had some trouble at the beginning designing how the "find records" query works. What we settled with meant that we didn't have to call `pageable.next()`, we only had to do this if we are in dry run mode.

I think because this is now doing strictly unit tests, mocking what you expect the db to return, instead of what we had before (a sort of mini integration test against an in memory db), this change in functionality has been missed.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/context/CommonBeans.java
```diff
@@ -47,32 +47,39 @@
...
-  @Value("${properties.apiary.queue-url}")
-  private String queueUrl;
+  @Autowired private List<SchedulerService> schedulerServices;
+  @Autowired private List<MessageEventHandler> handlers;
+  @Autowired private List<ListenerEventFilter> filters;
+
+  @Bean
+  public EnumMap<LifecycleEventType, SchedulerService> schedulerServiceMap(List<SchedulerService> schedulerServices) {
+    EnumMap<LifecycleEventType, SchedulerService> schedulerMap = new EnumMap<>(LifecycleEventType.class);
+    schedulerServices.stream().forEach(scheduler -> schedulerMap.put(scheduler.getLifecycleEventType(), scheduler));
```
max-jacobs: `schedulerServices` can directly call `forEach()` here

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/context/CommonBeans.java
```diff
@@ -47,32 +47,39 @@
 @EnableRetry(proxyTargetClass = true)
 public class CommonBeans {
 
-  @Value("${properties.apiary.queue-url}")
-  private String queueUrl;
+  @Autowired private List<SchedulerService> schedulerServices;
```
max-jacobs: Are these fields needed? If they are removed wont Spring just figure out what to inject from the method parameters?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
```diff
@@ -0,0 +1,147 @@
...
+
+public abstract class MessageEventHandler {
+
+  private static final Logger log = LoggerFactory.getLogger(MessageEventHandler.class);
+  private static final String CLIENT_ID = "apiary-metastore-event";
+  private final LifecycleEventType lifecycleEventType;
+  private final String cleanupDelay;
+  private final String hivePropertyKey;
+  private final List<FilterType> validFilterClasses;
+  @Autowired private EnumMap<FilterType, ListenerEventFilter> filterTypeMap;
```
max-jacobs: It feels weird to have this field autowired here, and not have this argument passed in from the subclass through the constructor. 

(we've had this discussion briefly before) If it was passed in from the subclass, it brings in to question the need for a mapping of  `ListenerEventFilter` and type altogether. The enums feel a bit superfluous as the classes themselves are pretty explicit. The list of valid filters is currently located in the subclass anyway, so this could just be a list of classes (the filters themselves are stateless, so we could remove the Spring bean annotations from them, and create them explicitly quite easily).

@massdosage any thoughts?

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
KenFigueiredo: IIRC my thought process was that since I had refactored `PagingCleanupService` to be more of an orchestrator of handlers, it didn't need to have that same scrutiny on the tests.

I do remember being slightly confused about the original query on how it progressed through in the default path w/o calling `next()`,so seeing that called out here means that I'd need to rethink this approach 😅.

Thinking about this a bit more for future proofing, if we're okay with converting the service into more of an orchestrator of handlers, we'd need to either consistently handle how we page through records at this level & refactor the call for `unreferenced` so it does need to call `next` or have it called out somewhere that any new lifecycle query has the expectation that it is written in a way that doesn't need to have `.next` called 

Completely open to suggestions on how to proceed on this part so it's not unintentionally breaking that original functionality.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
KenFigueiredo: agreed - the filter logic & sharing that logic between handlers via the parent hasn't been the cleanest solution. I think that if we are okay w/ having the filters be explicitly instantiated in the child handler class, that should simplify the setup and answer a couple of the other open questions I had about filters

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/context/CommonBeans.java
KenFigueiredo: nope! good catch, removed.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/handler/UnreferencedHandler.java
KenFigueiredo: As far as supporting further lifecycle changes - none that I can think of off the top of my head.

Good call out - Originally I had it from a generator and noticed that I removed it, my unit tests for the handler would fail w/ NPEs and forgot about it. Just realized that on save, if I don't have the setter my IDE will auto add a final modifier on to `private PathCleaner pathCleaner` effectively breaking the mock and the tests 🙃 

All's that to say is that  I updated my IDE settings and this is removed now...

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
max-jacobs: > I do remember being slightly confused about the original query on how it progressed through in the default path w/o calling next()

It progresses (if dryrun is off) because when a path is processed, its status is modified so "drops out" of the query result. Running the same query therefore will produce a different result each time without paging. The other case is where dryrun is enabled, and so no updates are made, so you have to do paging.

I imagine the `expired` handler will have to behave in a similar way? In which case, the distinction isn't about supporting the handlers, its about support dryruns. Does that make sense?

(I'm just thinking if there's possibly an edge case not covered by the original code: if an entire page of results are all originally `FAILED` and then _aren't_ updated because they fail again, I don't think we'll reach any more results...)

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
max-jacobs: Yeah I'm happy with either, if we have to maintain a list of filters for a handler, the least amount of code to do so would be nice. Let's see what others think...

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
KenFigueiredo: ah okay. That makes way more sense now, and see how that could potentially be problematic. Let me go ahead and try and setup a test for that & revert the next call for dry run / prod runs

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.0.0] - TBD
+### Added
+- Add `LifecycleEventType` enum in `beekeeper-core` to describe supported data Lifecycles.
+- Increase unit test coverage for `beekeeper-path-scheduler-apiary` package.
```
massdosage: Remove this and the next line, we generally don't mention changes to tests in the CHANGELOG as it's usually not that interesting. Most people reading it are downstream devs/users who are just interested in the changes to the application itself.

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
+- Increase coverage for integration tests.
+
+### Changed
+- Refactored internals of `beekeeper-path-scheduler-apiary` to support generic Lifecycle scheduling.
+    - Inserted additional workflow (handlers) between read & filter actions to support filters per Lifecycle type.
+    - MessageReaderAdapter now has additional logic to orchestrate the updated workflow.
+    - Rename `PathEvent` to `BeekeeperEvent` to better reflect event types.
```
massdosage:
```suggestion
    - Renamed `PathEvent` to `BeekeeperEvent` to better reflect event types.
```

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
+- Refactored internals of `beekeeper-path-scheduler` to support generic data Lifecycle scheduling.
+    - Rename `PathSchedulerService` to `UnreferencedPathSchedulerService` to differentiate types.
```
massdosage:
```suggestion
    - Renamed `PathSchedulerService` to `UnreferencedPathSchedulerService` to differentiate types.
```

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
+- Refactored internals of `beekeeper-cleanup` to support generic data Lifecycle deletions.
+    - Refactor `PagingCleanupService` to be a generic orchestrator of Lifecycle handlers.
```
massdosage:
```suggestion
    - Refactored `PagingCleanupService` to be a generic orchestrator of Lifecycle handlers.
```

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
+
```
massdosage: Above looks good, nice level of detail, thanks!

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/handler/UnreferencedHandler.java
```diff
@@ -0,0 +1,53 @@
...
+  private PathCleaner pathCleaner;
+
+  @Autowired
+  public UnreferencedHandler(@Qualifier("s3PathCleaner") PathCleaner pathCleaner) {
+    this.pathCleaner = pathCleaner;
+  }
+
+  @Override
+  public LifecycleEventType getLifecycleType() {
+    return EVENT_TYPE;
```
massdosage:
```suggestion
    return  LifecycleEventType.UNREFERENCED
```
No need for an intermediate variable to hold this.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/context/CommonBeansTest.java
```diff
@@ -54,18 +55,18 @@
...
-    System.clearProperty(AWS_S3_ENDPOINT_PROPERTY);
-  }
-
   @BeforeEach
   void setUp() {
     System.setProperty(AWS_REGION_PROPERTY, REGION);
     System.setProperty(AWS_S3_ENDPOINT_PROPERTY, ENDPOINT);
   }
 
+  @AfterAll
```
massdosage: Thanks for moving this, it reads better now!

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
massdosage: Yes please, IIRC we actually had a bug that our original tests didn't catch as we made some incorrect assumptions about how the paging worked. We then fleshed them out quite a bit to reproduce the bug and prevent future regressions. So whatever it makes sense to keep please do.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/EventTypeListenerEventFilter.java
```diff
@@ -23,11 +23,18 @@
...
 
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+
 @Component
 public class EventTypeListenerEventFilter implements ListenerEventFilter {
 
+  private final FilterType TYPE = FilterType.EVENT_TYPE;
+
+  @Override
+  public FilterType getFilterType() { return TYPE; }
```
massdosage:
```suggestion
  public FilterType getFilterType() { return  FilterType.EVENT_TYPE; }
```

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/MetadataOnlyListenerEventFilter.java
```diff
@@ -22,17 +22,24 @@
...
 
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+
 @Component
 public class MetadataOnlyListenerEventFilter implements ListenerEventFilter {
 
+  private final FilterType TYPE = FilterType.METADATA_ONLY;
+
+  @Override
+  public FilterType getFilterType() { return TYPE; }
```
massdosage:
```suggestion
  public FilterType getFilterType() { return  FilterType.METADATA_ONLY; }
```

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/TableParameterListenerEventFilter.java
```diff
@@ -21,17 +21,24 @@
...
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+
 @Component
 public class TableParameterListenerEventFilter implements ListenerEventFilter {
 
-  private static final String BEEKEEPER_TABLE_PARAMETER = "beekeeper.remove.unreferenced.data";
+  private final FilterType TYPE = FilterType.TABLE_PARAMETER;
+
+  @Override
+  public FilterType getFilterType() { return TYPE; }
```
massdosage:
```suggestion
  public FilterType getFilterType() { return FilterType.TABLE_PARAMETER; }
```

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/WhitelistedListenerEventFilter.java
```diff
@@ -23,13 +23,19 @@
...
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+
 @Component
 public class WhitelistedListenerEventFilter implements ListenerEventFilter {
 
   private static final String BEEKEEPER_HIVE_EVENT_WHITELIST = "beekeeper.hive.event.whitelist";
+  private final FilterType TYPE = FilterType.WHITELISTED;
+
+  @Override
+  public FilterType getFilterType() { return TYPE; }
```
massdosage:
```suggestion
  public FilterType getFilterType() { return FilterType.WHITELISTED; }
```

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/service/PagingCleanupServiceTest.java
KenFigueiredo: 👍 done. Reverted these tests to be more "integration-like" so it'll actually use the in memory db setup.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
KenFigueiredo: @max-jacobs @massdosage Updated this so that each handler is required to instantiate their own filters. This removes the enum & weird mapping to get this working correctly

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/handler/GenericHandlerTest.java
```diff
@@ -0,0 +1,85 @@
...
+  public void initTest() {
+    handler = new UnreferencedHandler(housekeepingPathRepository, pathCleaner);
+  }
+
+  @Test
+  public void typicalProcessDryRunPage() {
+    when(mockPage.getContent()).thenReturn(List.of(mockPath));
+    handler.processPage(mockPageable, mockPage, true);
+    verify(mockPageable).next();
+    verify(pathCleaner).cleanupPath(mockPath);
```
max-jacobs: Do we need to verify that `.next()` has been called here?

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/handler/UnreferencedHandlerTest.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Test
+  public void verifyLifecycle() {
+    assertThat(handler.getLifecycleType()).isEqualTo(LifecycleEventType.UNREFERENCED);
+  }
+
+  @Test
+  public void verifyHousekeepingPathFetch() {
+    handler.findRecordsToClean(null, null);
```
max-jacobs: These nulls may as well be something more meaningful (like "now" and an empty page), though it doesn't really change much.

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperPathSchedulerApiaryIntegrationTest.java
```diff
@@ -107,98 +116,213 @@ public static void init() throws IOException, SQLException {
...
+        "s3://unreferencedTableLocation",
+        "s3://partitionLocation",
+        "s3://unreferencedPartitionLocation",
+        false, false);
+    amazonSQS.sendMessage(sendMessageRequest(alterPartitionSqsMessage.getFormattedString()));
+    CreateTableSqsMessage createTableSqsMessage = new CreateTableSqsMessage(
+        "s3://unreferencedTableLocation", false, true);
+    amazonSQS.sendMessage(sendMessageRequest(createTableSqsMessage.getFormattedString()));
+
+    await().atMost(TIMEOUT, TimeUnit.SECONDS).until(() -> mySqlTestUtils.unreferencedRowsInTable(PATH_TABLE) == 0);
```
max-jacobs: I think this will resolve to true almost immediately (maybe before even the reader has a chance to poll). I think we can remove this test and place faith in our unit tests for this one.

max-jacobs: Same for the next two, anything which checks that there are zero rows in the table really - as it's difficult to know for sure that the message has been processed + rejected _before_ doing this check. Open to ideas for solving this one though.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/handler/GenericHandlerTest.java
KenFigueiredo: Given the context for dry runs and the setup we have I think it's a useful check to have. If we don't call next on the pageable during a dry run we'll always hit the same record right?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperPathSchedulerApiaryIntegrationTest.java
KenFigueiredo: Hmm - good point on this. I think a good way to test this would be to check is to do the await on when the SQS is empty. Once that's empty we have a reasonable assumption that beekeeper has popped & processed that message so we should be able to continue with the mysql check

KenFigueiredo: 🤔 I implemented it with what I had mentioned above. This could still lead to some race conditions as we aren't dictating when beekeeper is actually going to run it's action.

The only way I can see us setting this up to be more consistent is if we move the beekeeper execute startup in the `beforeEach` block into the tests themselves after we setup our queue to be what we expect.

Thoughts on that approach?

max-jacobs: Hmm, yes I don't think this solves the problem either, as we can still have a situation where the message is read, and then the test finishes, before Beekeeper does anything. Even if we start Beekeeper after the message has been put on the queue, as soon as the message is read we will have the same issue.

Originally, as with all integration tests, these were set up to test the happy path scenarios and ensure the application wires up correctly. I don't think we need to be this specific, and we can rely on our unit tests to ensure these paths are covered. In which case, I think we can remove these three. What do you think?

KenFigueiredo: works for me! good to have that context on how integration tests were setup originally.
Figured I'd take a crack at it at the very least, but went ahead and removed those tests.

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperPathSchedulerApiaryIntegrationTest.java
```diff
@@ -107,98 +114,182 @@ public static void init() throws IOException, SQLException {
...
   @Test
-  public void healthCheck() {
+  void healthCheck() {
     CloseableHttpClient client = HttpClientBuilder.create().build();
     HttpGet request = new HttpGet(HEALTHCHECK_URI);
     HttpCoreContext context = new HttpCoreContext();
-    await().atMost(30, TimeUnit.SECONDS)
-      .until(() -> client.execute(request, context).getStatusLine().getStatusCode() == 200);
+    await().atMost(TIMEOUT, TimeUnit.SECONDS)
+        .until(() -> client.execute(request, context).getStatusLine().getStatusCode() == 200);
```
massdosage: Indentation change looks wrong (4 instead of 2)

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/MetadataOnlyListenerEventFilter.java
```diff
@@ -22,17 +22,19 @@
...
   @Override
-  public boolean filter(ListenerEvent listenerEvent) {
+  public boolean filter(ListenerEvent listenerEvent, LifecycleEventType lifecycleEventType) {
     EventType eventType = listenerEvent.getEventType();
     switch (eventType) {
     case ALTER_PARTITION:
       AlterPartitionEvent alterPartitionEvent = (AlterPartitionEvent) listenerEvent;
       return isMetadataUpdate(alterPartitionEvent.getOldPartitionLocation(),
-        alterPartitionEvent.getPartitionLocation());
+          alterPartitionEvent.getPartitionLocation());
```
massdosage: Indentation

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/WhitelistedListenerEventFilter.java
```diff
@@ -39,9 +41,9 @@ public boolean filter(ListenerEvent listenerEvent) {
 
   private boolean isWhitelisted(ListenerEvent listenerEvent, String whitelist) {
     return Arrays.stream(whitelist.split(","))
-      .map(String::trim)
-      .anyMatch(whitelistedEvent -> whitelistedEvent.equalsIgnoreCase(listenerEvent.getEventType()
-        .toString()));
+        .map(String::trim)
```
massdosage: Indentations here and next 2 lines

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/filter/WhitelistedListenerEventFilterTest.java
```diff
@@ -73,8 +72,8 @@ public void filterNonDefaultEvents(EventType eventType) {
   public void filterWhitelistedEvent(String whitelist) {
     when(listenerEvent.getEventType()).thenReturn(EventType.DROP_TABLE);
     when(listenerEvent.getTableParameters())
-      .thenReturn(Map.of(BEEKEEPER_HIVE_EVENT_WHITELIST, whitelist));
-    boolean filter = listenerEventFilter.filter(listenerEvent);
+        .thenReturn(Map.of(BEEKEEPER_HIVE_EVENT_WHITELIST, whitelist));
```
massdosage: Indentation (and in methods below)

PR #37, opened 2020-01-17
Add Travis-CI Integration to Beekeeper

KenFigueiredo:
In order to keep a higher and cleaner standard of code on our OSS repositories, we should setup a Travis CI job to auto run & report the results & coverage of tests for a given open PR.

This brings Beekeeper inline with other OSS EG repositories and should allow for a higher level of confidence in any given change being merged.

Note: This won't currently run on this repository until merged, but here is the result in my fork: https://github.com/KenFigueiredo/beekeeper/pull/2


.travis.yml
```diff
@@ -0,0 +1,56 @@
...
+cache:
+  directories:
+    - ~/.m2/repository
+
+install:
+  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true --quiet --batch-mode --show-version --activate-profiles travis
+
+.base_unit_test: &base_unit_test
+  stage: unit_test
+  if: type = push
```
massdosage: Why do we need this?

.travis.yml
```diff
@@ -0,0 +1,56 @@
...
+  script:
+    - mvn -pl "$TEST_SUITE" test
+
+.base_integration_test: &base_integration_test
+  stage: integration_test
+  if: type = pull_request
+  script:
+    - mvn -pl 'beekeeper-integration-tests' -Dtest=$TEST_FILE test
+
+jobs:
```
massdosage: I think you could do all of the above using surefire and/or maven include/exclude rules for *IntegrationTest. Having said that, do we need this separation of unit and integration tests? Shouldn't we have the CI server run *everything* every time - package, coverage, checkstyle, findbugs and all the different tests? It's less to configure here and we don't mind if this takes a bit longer to run in the background (it can take a while to acquire and spin up VMs anyway).

.travis.yml
```diff
@@ -0,0 +1,56 @@
...
+
+.base_unit_test: &base_unit_test
+  stage: unit_test
+  if: type = push
+  script:
+    - mvn -pl "$TEST_SUITE" test
+
+.base_integration_test: &base_integration_test
+  stage: integration_test
+  if: type = pull_request
```
massdosage: Why not run on pushes and PRs?

.travis.yml
```diff
@@ -0,0 +1,56 @@
...
+  script:
+    - mvn -pl 'beekeeper-integration-tests' -Dtest=$TEST_FILE test
```
massdosage: If we do decide to keep this please use the verbose option for the CLI arguments (`--projects` in this case) as they're more readable.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.2.2] - TBD
+### Added
+- Added Travis-CI pipeline to verify Unit/Integration tests on push & PR
```
massdosage: Since this is an internal feature that end users don't see or care about it's one of the rare cases that we don't update the changelog.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.2.2] - TBD
```
massdosage: 1.2.1 would need to be released before we added a 1.2.2 (so if we were going to document this it should go under 1.2.1).

.travis.yml
```diff
@@ -0,0 +1,56 @@
...
+
+language: java
+
+services:
+  - docker
+
+jdk:
+  - openjdk11
+
+os:
```
massdosage: Can you put OS at the top please? I like to start with the more general stuff and then move down to more specific.

.travis.yml
KenFigueiredo: Answering this and the comment for `if: type = push`, `if: type = pull_request` here.

Essentially the way I have this pipeline setup is to run the full suite of unit tests on push, but only run integration tests on open PRs.  I've got all the unit test suites broken up as well to run them in separate processes in Travis so they'll report failures individually. Overall It's more of a choice to keep build times for commits quicker so if you're developing a feature you'll have a faster feedback loop to see that the change you made has broken something before you open up a PR. 

Generally I've seen in practice teams do integration testing on only PR events or after merging to master (as part of an automated deployment setup) since integration tests usually are a large factor of time slower than unit tests.

Although with all that said - it really comes down to preference of what workflow works best for our team, so if we want everything to just run on every commit that is totally fine too 😄

.travis.yml
KenFigueiredo: see other comment

.travis.yml
KenFigueiredo: see other comment

.travis.yml
massdosage: For dev feedback I recommend running the unit tests locally. I think CI should run everything and be more of a safety net when you're asking others to review your code, which is usually a bunch of commits rolled up into a push. I definitely have a preference to just run the whole Maven build each time this happens as that's what we've been doing up to now in other projects and it seems to work well and everyone is happy with that. Speak up if anyone else disagrees ;)

KenFigueiredo: Reverting that now to always run the full test suite on commit 🚀 

It looks like I'm running into some issues w/ the integration suite now though where it's not completing some tests within the allotted time which is a bit strange. Looking into that some more now

massdosage: Let's see what Travis thinks of this, we can always update this later...

PR #38, opened 2020-01-22
Get tests running in Eclipse and fix Travis

massdosage:
- upgrade junit to 5.6.0 (and import bom)
- upgrade mockito to 3.2.4
- removed override of surefire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [1.2.1] - TBD
 ### Changed
 - `S3Client.listObjects()` to list all objects at a key with batching.  
+- `eg-oss-parent` version updated to 1.3.1 (was 1.1.0).
 
 ## [1.2.0] - 2020-01-06
 ### Adding
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>com.expediagroup</groupId>
     <artifactId>eg-oss-parent</artifactId>
-    <version>1.1.0</version>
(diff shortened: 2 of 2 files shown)


massdosage:
I've purposefully not updated the CHANGELOG since these are internal-only changes.

massdosage:
Looks like we need to enable surefire specifically, will do this in eg-oss-parent and upgrade that and then depend on the new version from here.

massdosage:
OK, tests run fine for me via Maven CLI locally and in Eclipse. We're getting timeouts running the integration in Travis but this could be due to general Travis scaling issues which we've seen before. @max-jacobs would you mind trying this branch out locally and just checking it works like you would expect?

KenFigueiredo:
Hey @massdosage!

I opened up https://github.com/ExpediaGroup/beekeeper/pull/39 against this branch to see if the issues went away in GH Actions, which they seem to do. I've set that pipeline up so if we want to try rolling either both Travis & GH Actions on this repository or swap out to just GH actions to have more consistent builds 

Let me know if you have any additional thoughts / concerns about that approach 😄

massdosage:
> Hey @massdosage!
> 
> I opened up #39 against this branch to see if the issues went away in GH Actions, which they seem to do. I've set that pipeline up so if we want to try rolling either both Travis & GH Actions on this repository or swap out to just GH actions to have more consistent builds
> 
> Let me know if you have any additional thoughts / concerns about that approach smile

Excellent, I'll take a look at that when I get a chance. For now let's go with GH Actions *and* Travis and we can compare them over the next few weeks. If GH actions proves to be more stable (which is what I suspect will be the case) then we can remove Travis.

PR #41, opened 2020-01-24
trim github action build down

massdosage:
From looking at the GH actions output I think we can remove some of the steps. I don't think the cache is working and we probably don't really need it if we only run one action. I think we inherited the "install" step from how we did things in Travis but I don't think we need it. I also removed the timeout as I'd rather use GH's default and only specifically configure values for things if we need to.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @ExpediaGroup/beekeeper-commiters
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,26 +4,13 @@ on: [push, pull_request]
 
 jobs:
   test:
-    name: Run all tests
+    name: Package and run all tests
     runs-on: ubuntu-18.04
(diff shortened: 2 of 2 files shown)


massdosage:
@KenFigueiredo any thoughts? How did you test these? I'm guessing the GH action run above against this PR will use the main.yml from master and not my changed version in the branch right?

KenFigueiredo:
@massdosage No strong thoughts here really. If the cache isn't working as expected then you're right in that it's probably not worth pulling in. We may want to revisit this at a future date if the build ends up taking longer than we want to see if we can reduce the time of the maven install step.

As far as testing - the way I understood the document is that since this branch is directly on this repo any changes that you make to the github/actions.yml will be ran for this branch. 

Fork github/action modifications are ignored though as it presents a security risk depending on what your job does, so they always take the target pr branch.

massdosage:
Yeah, it looks like it used my changes and it all passed in GH so we're good.

PR #48, opened 2020-04-21
Adding logo

max-jacobs:


README.md
```diff
@@ -1,3 +1,5 @@
+![Logo](.README_images/full_logo.png)
```
patduin, line 1: why you need all the other added images?

max-jacobs: for safe keeping I guess

patduin: ok :)

PR #49, opened 2020-04-23
Add vacuum tool

massdosage:
This PR adds the beekeeper-vacuum-tool as a sub-module. It's a "port" of code that we previously had internally due to its dependence on a fork of Hive which is no longer required since Hive 2.3.7 was released. Noteable changes in this PR:
* Fixed an issue I found as a side-effect where Hibernate validation was turning off null-checks for JPA entities (see https://www.baeldung.com/hibernate-notnull-vs-nullable). This required moving some Spring dependencies around and excluding `hibernate-validator`.
* Tidied up dependency order in all the poms to be alphabetical (might look like overkill but I was having problems keeping track of Spring dependencies in use as they were scattered within each pom).


beekeeper-vacuum-tool/pom.xml
```diff
@@ -0,0 +1,245 @@
...
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
max-jacobs, line 182: Can we drop this plugin?

massdosage: Uncertain, I just copied it over from before, let me see what happens if removed...

massdosage: Looks like it can go, I'll remove shortly.

massdosage:
> Looks pretty good, I wonder if its worth checking now if there are any problems with running the Path Scheduler on an EMR cluster - just thinking ahead to integrating with Dronefly, or if we should save that for another day

What exactly would you want to check and how? I'm happy to try some more real world tests but might need some help performing them etc. I would like to do a release of Beekeeper once this is merged so we should be confident that it all works as expected.

max-jacobs:
Thinking about this, Path Scheduler Apiary wont exist anymore in a post Dronefly world, so actually I'm not sure what we would run. Never mind, something to think about later I think.

PR #51, opened 2020-06-18
Add Time To Live (TTL) for tables

vedantchokshi:
Fixes #


README.md
```diff
@@ -28,8 +28,8 @@ Beekeeper comprises two separate Spring-based Java applications. One application
 2. An operation is executed on the table that orphans some data (alter partition, drop partition, etc.)
 3. Hive Metastore events are emitted by the [Hive Metastore Listener](https://github.com/ExpediaGroup/apiary-extensions/tree/master/apiary-metastore-listener) as a result of the operation.
 4. Hive events are picked up from the queue by Beekeeper using [Apiary Receiver](https://github.com/ExpediaGroup/apiary-extensions/tree/master/apiary-receivers).
-5. Beekeeper processes these messages and schedules orphaned paths for deletion by adding them to a database.
-6. The scheduled paths are deleted by Beekeeper after a configurable delay, the default is 3 days (see [Hive table configuration](#hive-table-configuration) for more details.)
+5. Beekeeper processes these messages and schedules orphaned paths or expired tables for deletion by adding them to a database.
+6. The scheduled paths or tables are deleted by Beekeeper after a configurable delay, the default is 3 days for tables and 30 days for tables (see [Hive table configuration](#hive-table-configuration) for more details.)
```
JayGreeeen, line 32: `3 days for tables and 30 days for tables` ?

README.md
```diff
@@ -168,6 +170,7 @@ By default, `beekeeper-path-scheduler-apiary` listens on port 8080 and `beekeepe
 | `apiary.queue-url`                  | Yes      | URL for SQS queue. |
 | `apiary.cleanup-delay-property-key` | No       | Table parameter to use for Apiary listener. Default value is `beekeeper.unreferenced.data.retention.period`. |
 | `beekeeper.default-cleanup-delay`   | No       | Default Time To Live (TTL) for orphaned paths in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) format: only days, hours, minutes and seconds can be specified in the expression. Default value is `P3D` (3 days). |
+| `beekeeper.default-expiration-delay`| No    | Default Time To Live (TTL) for tables in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) format: only days, hours, minutes and seconds can be specified in the expression. Default value is `P3D` (3 days). |
```
JayGreeeen, line 173: Default is 30 days so should be P30D

README.md
vedantchokshi: Whoops I mean't to write paths

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/EntityHousekeepingPath.java
```diff
@@ -52,7 +52,7 @@
 
   @Column(name = "path_status", nullable = false)
   @Enumerated(EnumType.STRING)
-  private PathStatus pathStatus;
+  private HousekeepingStatus pathStatus;
```
max-jacobs, line 55: Was the field name changed to match the interface? The column annotation will map the field to the correct column (`path_status`) in the table, but it does look a bit weird. Perhaps we could run a table migration to alter the name of this column from `path_status` to `housekeeping_status`?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/repository/HousekeepingTableRepository.java
```diff
@@ -0,0 +1,43 @@
...
+      + "and (t.housekeepingStatus = 'SCHEDULED' or t.housekeepingStatus = 'FAILED') "
+      + "and t.modifiedTimestamp <= :instant order by t.modifiedTimestamp")
+  Page<EntityHousekeepingTable> findRecordsForCleanupByModifiedTimestamp(@Param("instant") LocalDateTime instant,
+      Pageable pageable);
+
+  @Query(value = "select t from EntityHousekeepingTable t "
+      + "where t.databaseName = :databaseName "
+      + "and t.tableName = :tableName "
+      + "and (t.housekeepingStatus = 'SCHEDULED' or t.housekeepingStatus = 'FAILED')")
+  Optional<EntityHousekeepingTable> findRecordForCleanupByDatabaseAndTable(@Param("databaseName") String databaseName,
```
max-jacobs, line 41: I'm not sure you need to have the query annotation here. I think `findByDatabaseNameAndTableName(...)` will do what you need it to do automatically.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/LocationOnlyUpdateListenerEventFilter.java
```diff
@@ -25,25 +25,25 @@
 import com.expediagroup.beekeeper.core.model.LifecycleEventType;
 
 @Component
-public class MetadataOnlyListenerEventFilter implements ListenerEventFilter {
+public class LocationOnlyUpdateListenerEventFilter implements ListenerEventFilter {
```
max-jacobs, line 28: I think maybe lets stick to the old name? It matches a similar filter I think that's in Shunting Yard. What do you think?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/WhitelistedListenerEventFilter.java
```diff
@@ -31,10 +31,12 @@
   private static final String BEEKEEPER_HIVE_EVENT_WHITELIST = "beekeeper.hive.event.whitelist";
 
   @Override
-  public boolean filter(ListenerEvent listenerEvent, LifecycleEventType lifecycleEventType) {
+  public boolean isFilteredOut(ListenerEvent listenerEvent, LifecycleEventType lifecycleEventType) {
     Map<String, String> tableParameters = listenerEvent.getTableParameters();
-    if (tableParameters != null && tableParameters.get(BEEKEEPER_HIVE_EVENT_WHITELIST) != null) {
-      return !isWhitelisted(listenerEvent, tableParameters.get(BEEKEEPER_HIVE_EVENT_WHITELIST));
+    if (tableParameters != null
+        && tableParameters.get(WhitelistedListenerEventFilter.BEEKEEPER_HIVE_EVENT_WHITELIST) != null) {
```
max-jacobs, line 37: This can go: `WhitelistedListenerEventFilter.` and the same on line 39.

I think also we should rename this class and the table parameter. Perhaps we can do it as part of another ticket, but we should no longer make references to a "whitelist". Let's call this "allowlist" instead.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/ExpiredMessageHandler.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Autowired
+  public ExpiredMessageHandler(
+      @Value("${properties.beekeeper.default-expiration-delay}") String cleanupDelay
+  ) {
+    super(cleanupDelay, EXPIRED_DATA_RETENTION_PERIOD_PROPERTY_KEY, LIFECYCLE_EVENT_TYPE);
+    this.filters = List.of(
+        new EventTypeListenerEventFilter(EVENT_CLASSES),
+        new TableParameterListenerEventFilter(),
+        new WhitelistedListenerEventFilter());
```
max-jacobs, line 59: Why do we need this filter?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/ExpiredMessageHandler.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Override
+  protected List<ExpiredEventModel> generateEventModels(ListenerEvent event) {
+    List<ExpiredEventModel> eventPaths = new ArrayList<>();
+
+    switch (event.getEventType()) {
+    case ALTER_TABLE:
+      eventPaths.add(new ExpiredEventModel(LIFECYCLE_EVENT_TYPE, SCHEDULED));
+      break;
```
max-jacobs, line 97: We could have a default setting here that throws an exception as we shouldn't expect to reach it

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/UnreferencedMessageHandler.java
```diff
@@ -30,64 +33,86 @@
...
 import com.expediagroup.beekeeper.scheduler.apiary.filter.WhitelistedListenerEventFilter;
-import com.expediagroup.beekeeper.scheduler.apiary.model.EventModel;
+import com.expediagroup.beekeeper.scheduler.apiary.model.UnreferencedEventModel;
 
 @Component
-public class UnreferencedMessageHandler extends MessageEventHandler {
+public class UnreferencedMessageHandler extends MessageEventHandler<EntityHousekeepingPath, UnreferencedEventModel> {
 
   private static final Logger log = LoggerFactory.getLogger(UnreferencedMessageHandler.class);
+  private static final String UNREFERENCED_DATA_RETENTION_PERIOD_PROPERTY_KEY = "beekeeper.unreferenced.data.retention.period";
```
max-jacobs, line 49: Where did this annotation used to live?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/model/ExpiredEventModel.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.expediagroup.beekeeper.core.model.HousekeepingStatus;
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+
+public class ExpiredEventModel implements EventModel {
+
+  private final LifecycleEventType lifecycleEventType;
+  private final HousekeepingStatus housekeepingStatus;
+
+  public ExpiredEventModel(LifecycleEventType lifecycleEvent, HousekeepingStatus housekeepingStatus) {
```
max-jacobs, line 26: Why does the `ExpiredEventModel` need the `housekeepingStatus`? Surely that's going to be `SCHEDULED`? Or am I missing something?

beekeeper-path-scheduler-apiary/src/main/resources/beekeeper-path-scheduler-apiary-application.yml
```diff
@@ -13,5 +13,4 @@ spring:
 properties:
   beekeeper:
     default-cleanup-delay: P3D
-  apiary:
-    cleanup-delay-property-key: beekeeper.unreferenced.data.retention.period
+    default-expiration-delay: P30D
```
max-jacobs, line 16: Ah so this is where the key used to come from? I think that's a good move as we're not going to change this key now.

README.md
```diff
@@ -168,6 +170,7 @@ By default, `beekeeper-path-scheduler-apiary` listens on port 8080 and `beekeepe
 | `apiary.queue-url`                  | Yes      | URL for SQS queue. |
 | `apiary.cleanup-delay-property-key` | No       | Table parameter to use for Apiary listener. Default value is `beekeeper.unreferenced.data.retention.period`. |
```
max-jacobs, line 171: I think this needs to be removed as its now hardcoded

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/filter/EventTypeListenerEventFilterTest.java
```diff
@@ -46,36 +57,36 @@
 
   @Test
   public void typicalFilterAlterPartitionEvent() {
-    when(alterPartitionEvent.getEventType()).thenReturn(EventType.ALTER_PARTITION);
-    boolean filter = listenerEventFilter.filter(alterPartitionEvent, UNREFERENCED);
+    when(this.alterPartitionEvent.getEventType()).thenReturn(ALTER_PARTITION);
```
max-jacobs, line 60: Don't think you need all the `this`

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/filter/TableParameterListenerEventFilterTest.java
```diff
@@ -43,41 +43,44 @@
 
   @Test
   public void typicalFilterBeekeeperManaged() {
-    when(alterPartitionEvent.getTableParameters())
-        .thenReturn(Map.of(BEEKEEPER_TABLE_PARAMETER, BEEKEEPER_MANAGED));
-    Boolean filter = listenerEventFilter.filter(alterPartitionEvent, UNREFERENCED);
+    when(this.alterPartitionEvent.getTableParameters())
```
max-jacobs, line 46: Same in this class with the `this`

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/filter/WhitelistedListenerEventFilterTest.java
```diff
@@ -47,17 +47,17 @@
   @EnumSource(value = EventType.class, names = { "ALTER_PARTITION",
                                                  "ALTER_TABLE" })
   public void filterDefaultEvents(EventType eventType) {
-    when(listenerEvent.getEventType()).thenReturn(eventType);
-    boolean filter = listenerEventFilter.filter(listenerEvent, UNREFERENCED);
+    when(this.listenerEvent.getEventType()).thenReturn(eventType);
```
max-jacobs, line 50: And here, is this Intellij doing formatting?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/EntityHousekeepingPath.java
patduin: I'm not too bothered if it stays pathStatus in the table what do others think?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/EntityHousekeepingTable.java
```diff
@@ -0,0 +1,290 @@
...
+import javax.persistence.Table;
+
+import org.hibernate.annotations.UpdateTimestamp;
+
+import com.expediagroup.beekeeper.core.error.BeekeeperException;
+import com.expediagroup.beekeeper.core.monitoring.MetricTag;
+
+@Entity
+@Table(name = "table_housekeeping")
+public class EntityHousekeepingTable implements Housekeeping {
```
patduin, line 38: I think maybe we can call this EntityHousekeepingMetadata
And add a column: partitionValue String, which will be null if it is the table? thoughts?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.2.0] - TBD
```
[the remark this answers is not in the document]
vedantchokshi: I thought so but then I wasn't sure because its doesn't really include changes that aren't backwards compatible. @patduin @abhimanyugupta07 @JayGreeeen what do you guys think?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/EntityHousekeepingPath.java
massdosage: I'd be up for renaming the column if it's not too much effort to roll that out.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/EntityHousekeepingPath.java
```diff
@@ -35,7 +35,7 @@
 
 @Entity
 @Table(name = "path")
-public class EntityHousekeepingPath implements HousekeepingPath {
+public class EntityHousekeepingPath implements Housekeeping {
```
massdosage, line 38: Could we now rename this to `HousekeepingPath` since the interface now has a different name? That might also require a table name change?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/EntityHousekeepingTable.java
massdosage: Agree on not having the name "table" in it. I think we can also remove `Entity`, we only did that for the previous class to not clash with the `HousekeepingPath` interface name. `HouseKeepingMetadata` sounds good?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/Housekeeping.java
```diff
@@ -20,18 +20,14 @@
 
 import com.expediagroup.beekeeper.core.monitoring.Taggable;
 
-public interface HousekeepingPath extends Taggable {
+public interface Housekeeping extends Taggable {
```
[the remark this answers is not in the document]
massdosage: `HouseKeepingEntity` is OK, or maybe `HousekeepingItem` or `HousekeepingEntry`? Might need to run a poll and get everyone to vote ;)

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/EventTypeListenerEventFilter.java
```diff
@@ -15,25 +15,26 @@
...
 
+  private final List<Class<? extends ListenerEvent>> eventClasses;
+
+  public EventTypeListenerEventFilter(List<Class<? extends ListenerEvent>> eventClasses) {
+    this.eventClasses = eventClasses;
+  }
+
   @Override
-  public boolean filter(ListenerEvent listenerEvent, LifecycleEventType lifecycleEventType) {
+  public boolean isFilteredOut(ListenerEvent listenerEvent, LifecycleEventType lifecycleEventType) {
```
massdosage, line 36:
```suggestion
  public boolean isFiltered(ListenerEvent listenerEvent, LifecycleEventType lifecycleEventType) {
```

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/WhitelistedListenerEventFilter.java
massdosage: +1 for renaming to AllowList

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/filter/WhitelistedListenerEventFilterTest.java
massdosage: Make it sttttttop! ;P

CHANGELOG.md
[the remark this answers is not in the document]
abhimanyugupta07: I am fine with this being a major update. This change is big enough.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/UnreferencedMessageHandler.java
```diff
@@ -30,64 +33,86 @@
...
+        .databaseName(listenerEvent.getDbName())
+        .path(eventModel.getCleanupPath());
+
+    return builder.build();
+  }
 
   @Override
-  protected List<EventModel> generateEventModels(ListenerEvent event) {
-    List<EventModel> eventPaths = new ArrayList<>();
+  protected List<UnreferencedEventModel> generateEventModels(ListenerEvent event) {
```
patduin, line 99: why is this returning a list we never return more than 1???

patduin: ah it can be empty list got it.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
```diff
@@ -53,23 +50,25 @@
...
       return Collections.emptyList();
     }
 
-    return generateHouseKeepingPaths(listenerEvent);
+    return generateHousekeepingEntities(listenerEvent);
   }
 
-  abstract List<EventModel> generateEventModels(ListenerEvent listenerEvent);
-
   protected abstract List<ListenerEventFilter> getFilters();
```
patduin, line 63: This could perhaps be set in the constructor?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
```diff
@@ -78,37 +77,22 @@ private boolean shouldFilterMessage(ListenerEvent listenerEvent) {
    * @param listenerEvent Listener event from the current message
    * @return list of housekeeping paths. This can be an empty list if there are no valid paths.
    */
-  private List<HousekeepingPath> generateHouseKeepingPaths(ListenerEvent listenerEvent) {
+  private List<Housekeeping> generateHousekeepingEntities(ListenerEvent listenerEvent) {
```
patduin, line 80: this method + the two abstract methods it calls are complex and flow is hard to follow, the generics don't help. EventModel implementations aren't really used anywhere except in the concrete subclasses and are unnecessary maybe. Perhaps we can instantiate this class with HouseKeepingEntityGenerator class that can turn do ```List< Housekeeping> generate(listenerEvent)```. Encapsulating the two abstract methods which can be removed.
Then can have two implementations (similar to the two subclasses we have now) that way we move the inheritance to composition which will be easier to test.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/model/EventModel.java
```diff
@@ -17,21 +17,7 @@
 
 import com.expediagroup.beekeeper.core.model.LifecycleEventType;
 
-public class EventModel {
+public interface EventModel {
```
patduin, line 20: I think this can problably go completely, see https://github.com/ExpediaGroup/beekeeper/pull/51/files#r444041998

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/EntityHousekeepingTable.java
vedantchokshi: I will add the partition value in the next set of changes. I didn't want to add the partition stuff in this PR because it was already getting too big. Is that okay?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/Housekeeping.java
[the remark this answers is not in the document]
vedantchokshi: I've just used HousekeepingEntity for now. @max-jacobs @abhimanyugupta07 @massdosage @JayGreeeen any other preferences?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/repository/HousekeepingTableRepository.java
vedantchokshi: Will it do the filtering by `SCHEDULED` or `FAILED`?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/EventTypeListenerEventFilter.java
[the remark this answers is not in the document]
vedantchokshi: True. I think Adrian's suggestion is better. What do you think?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/LocationOnlyUpdateListenerEventFilter.java
vedantchokshi: I also think the name is a lot clearer. I think in the context of this project, this name makes sense. Is it important for the names don't match?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/ExpiredMessageHandler.java
vedantchokshi: My bad. I misunderstood this so I'll remove it

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/ExpiredMessageHandler.java
vedantchokshi: Well, all the other events will throw that exception then right? Shouldn't we just ignore events we don't need and only care about the ones we want?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
vedantchokshi: Yeah we could but then it makes a bit mess when extend the class because the `super()` method has to be called first in the parameter. So creating the list instance in the super call will make it a bit messy. Example:

This doesn't work
```
  @Autowired
  public ExpiredMessageHandler(
      @Value("${properties.beekeeper.default-expiration-delay}") String cleanupDelay
  ) {
    List<ListenerEventFilter> filters = List.of(
        new EventTypeListenerEventFilter(EVENT_CLASSES),
        new TableParameterListenerEventFilter());
    super(cleanupDelay, EXPIRED_DATA_RETENTION_PERIOD_PROPERTY_KEY, LIFECYCLE_EVENT_TYPE, filters);
  }
```

This will get messy if you have lots of filters:
```
  @Autowired
  public ExpiredMessageHandler(
      @Value("${properties.beekeeper.default-expiration-delay}") String cleanupDelay
  ) {
    super(cleanupDelay, EXPIRED_DATA_RETENTION_PERIOD_PROPERTY_KEY, LIFECYCLE_EVENT_TYPE, List.of(
        new EventTypeListenerEventFilter(EVENT_CLASSES),
        new TableParameterListenerEventFilter()));
  }
```

What do you think?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
```diff
@@ -78,37 +77,22 @@ private boolean shouldFilterMessage(ListenerEvent listenerEvent) {
...
-        .creationTimestamp(LocalDateTime.now())
-        .cleanupDelay(extractCleanupDelay(listenerEvent))
-        .lifecycleType(event.getLifecycleEvent().name())
-        .clientId(CLIENT_ID)
-        .tableName(listenerEvent.getTableName())
-        .databaseName(listenerEvent.getDbName())
-        .path(event.getCleanupPath());
-
-    return builder.build();
+  private LifecycleEventType getLifecycleEventType() {
```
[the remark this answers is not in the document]
vedantchokshi: Yeah. I think it was just there just in case

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/EventTypeListenerEventFilter.java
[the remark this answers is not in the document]
patduin: sure

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
patduin:
```
private final static List<> FILTERS = List.of(
        new EventTypeListenerEventFilter(EVENT_CLASSES),
        new TableParameterListenerEventFilter());

@Autowired
  public ExpiredMessageHandler(
      @Value("${properties.beekeeper.default-expiration-delay}") String cleanupDelay
  ) {
    super(cleanupDelay, EXPIRED_DATA_RETENTION_PERIOD_PROPERTY_KEY, LIFECYCLE_EVENT_TYPE, FILTERS);
  }
```

maybe?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/repository/HousekeepingTableRepository.java
max-jacobs: Ah fair enough, didn't see that

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/LocationOnlyUpdateListenerEventFilter.java
max-jacobs: Let's stick with it

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/WhitelistedListenerEventFilter.java
max-jacobs: Will raise a ticket

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/ExpiredMessageHandler.java
max-jacobs: Hmm, I think for the other flow we decided that if an event lands here which doesn't match one of the cases, something has gone wrong i.e. the filters should have filtered it out. So if we want only alter tables to get to this point, everything else will have to be filtered somehow.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/Housekeeping.java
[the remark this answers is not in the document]
max-jacobs: Works for me

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/service/PathSchedulerApiary.java
```diff
@@ -50,21 +50,21 @@ public PathSchedulerApiary(
 
   @Transactional
   public void scheduleBeekeeperEvent() {
-    Optional<BeekeeperEvent> pathToBeScheduled = beekeeperEventReader.read();
-    if (pathToBeScheduled.isEmpty()) { return; }
-    BeekeeperEvent beekeeperEvent = pathToBeScheduled.get();
-    List<HousekeepingPath> paths = beekeeperEvent.getHousekeepingPaths();
+    Optional<BeekeeperEvent> housekeepingEntitiesToBeScheduled = beekeeperEventReader.read();
```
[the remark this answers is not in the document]
max-jacobs: Sure thing

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/ExpiredMessageHandler.java
vedantchokshi: Ah yeah you're right. Sure I'll do that :-)

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/MessageEventHandler.java
vedantchokshi: Not sure if that would work because we insert mock filter objects with this other constructor for tests:

```
  public ExpiredMessageHandler(
      @Value("${properties.beekeeper.default-expiration-delay}") String cleanupDelay,
      List<ListenerEventFilter> filters
  ) {
    super(cleanupDelay, EXPIRED_DATA_RETENTION_PERIOD_PROPERTY_KEY, LIFECYCLE_EVENT_TYPE);
    this.filters = filters;
  }
```

Any other ideas?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/UnreferencedMessageHandler.java
vedantchokshi: This used to be in the config yaml file. Since its not supposed to be configurable, I moved it. Is that ok?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/model/ExpiredEventModel.java
vedantchokshi: Yup my bad. Removed it now

beekeeper-path-scheduler-apiary/src/main/resources/beekeeper-path-scheduler-apiary-application.yml
vedantchokshi: Yup :-)

beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/filter/EventTypeListenerEventFilterTest.java
vedantchokshi: Yh sorry. It was this plugin I had which automatically changed these files when I opened them. I have disabled it now though

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/Housekeeping.java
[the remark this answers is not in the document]
massdosage: I think I can live with that ;)

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/service/PathSchedulerApiary.java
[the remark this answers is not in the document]
massdosage: Well, it was a "path scheduler" so we should try call it a "X scheduler" instead. `HousekeepingScheduler`? `BeekeeperScheduler`? `ApiaryScheduler`?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/UnreferencedMessageHandler.java
```diff
@@ -30,64 +33,86 @@
...
-import com.expediagroup.beekeeper.scheduler.apiary.filter.MetadataOnlyListenerEventFilter;
+import com.expediagroup.beekeeper.scheduler.apiary.filter.LocationOnlyUpdateListenerEventFilter;
 import com.expediagroup.beekeeper.scheduler.apiary.filter.TableParameterListenerEventFilter;
 import com.expediagroup.beekeeper.scheduler.apiary.filter.WhitelistedListenerEventFilter;
-import com.expediagroup.beekeeper.scheduler.apiary.model.EventModel;
+import com.expediagroup.beekeeper.scheduler.apiary.model.UnreferencedEventModel;
 
 @Component
-public class UnreferencedMessageHandler extends MessageEventHandler {
+public class UnreferencedMessageHandler extends MessageEventHandler<EntityHousekeepingPath, UnreferencedEventModel> {
```
patduin, line 46: Sorry, I think we should renamed this: UnreferencedLocationMessageHandler?
So make a distinction that it's dealing with Location data not metadata,

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/handler/ExpiredMessageHandler.java
```diff
@@ -0,0 +1,102 @@
...
+import com.expediagroup.beekeeper.core.model.EntityHousekeepingTable;
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+import com.expediagroup.beekeeper.scheduler.apiary.filter.EventTypeListenerEventFilter;
+import com.expediagroup.beekeeper.scheduler.apiary.filter.ListenerEventFilter;
+import com.expediagroup.beekeeper.scheduler.apiary.filter.TableParameterListenerEventFilter;
+import com.expediagroup.beekeeper.scheduler.apiary.filter.WhitelistedListenerEventFilter;
+import com.expediagroup.beekeeper.scheduler.apiary.model.ExpiredEventModel;
+
+@Component
+public class ExpiredMessageHandler extends MessageEventHandler<EntityHousekeepingTable, ExpiredEventModel> {
```
patduin, line 42: rename maybe: ExpiredMetadataMessageHandler?
I feel it should be more clear what classes deal with data (locations) and which with metadata.

We potentially could have expired Locations or unreferenced metadata(? maybe not :))).

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/service/PathSchedulerApiary.java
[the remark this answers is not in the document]
vedantchokshi: So there was a package called `beekeeper-path-scheduler` which I have renamed to `beekeeper-scheduler`. To keep up with the current conventions, I renamed `beekeeper-path-scheduler-apiary` to `beekeeper-scheduler-apiary` which resulted in naming `PathSchedulerApiary` to `SchedulerApiary`. I think `ApiaryScheduler` sounds better but then how do we want to name the packages?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingEntity.java
```diff
@@ -0,0 +1,66 @@
...
+import java.time.Duration;
+import java.time.LocalDateTime;
+
+import com.expediagroup.beekeeper.core.monitoring.Taggable;
+
+public interface HousekeepingEntity extends Taggable {
+
+  String getLifecycleType();
+
+  void setLifecycleType(String lifecycleType);
```
patduin, line 27: Do you really need all those setters, you'll always have to build the concrete class right, this should probably just be a subset of the getters that are shared in concrete subclasses.

README.md
```diff
@@ -49,7 +49,9 @@ Beekeeper only actions on events which are marked with a specific parameter. Thi
...
-| `beekeeper.hive.event.whitelist=X` | No | Comma separated list of event types to whitelist. Valid event values are: `alter_partition`, `alter_table`, `drop_table`, `drop_partition`. | Beekeeper will only process whitelisted events. Default value: `alter_partition`, `alter_table`. |
+| `beekeeper.hive.event.whitelist=X` | No | Comma separated list of event types to whitelist for orphaned data. Valid event values are: `alter_partition`, `alter_table`, `drop_table`, `drop_partition`. | Beekeeper will only process whitelisted events. Default value: `alter_partition`, `alter_table`. |
+| `beekeeper.remove.expired.data=true`   | Yes |  `true` or `false`       | Set this parameter to enable TTL on your table. |
+| `beekeeper.expired.data.retention.period=X` | No | e.g. `P7D` or `PT3H` (based on [ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601)) | Set this parameter to set the TTL duration for your table. If this is either not set, or configured incorrectly, the default will be used. Default is 30 days. |
```
massdosage, line 54:
```suggestion
| `beekeeper.expired.data.retention.period=X` | No | e.g. `P7D` or `PT3H` (based on [ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601)) | Set this parameter to set the TTL duration for your table. If this is either not set, or configured incorrectly, the default will be used. Default value is `P30D` (30 days).|
```

CHANGELOG.md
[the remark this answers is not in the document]
massdosage: Let's make it so then ;)

beekeeper-scheduler-apiary/src/main/resources/db/migration/V2_1__Create_unpartitioned_table_housekeeping_table_and_alter_path_table_name_and_rename_path_status_column.sql
```diff
@@ -0,0 +1,19 @@
+USE beekeeper;
```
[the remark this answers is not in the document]
massdosage: I guess it's fine to have it in `resources` instead of `src/main/sql`

beekeeper-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/apiary/generator/HousekeepingEntityGeneratorTestBase.java
```diff
@@ -0,0 +1,75 @@
...
+import org.mockito.Mock;
+
+import com.expedia.apiary.extensions.receiver.common.event.EventType;
+import com.expedia.apiary.extensions.receiver.common.event.ListenerEvent;
+
+import com.expediagroup.beekeeper.core.model.HousekeepingEntity;
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+import com.expediagroup.beekeeper.scheduler.apiary.generator.utils.CleanupDelayExtractor;
+
+public class HousekeepingEntityGeneratorTestBase {
```
massdosage, line 38: Might as well make it `abstract` as I don't think one would ever instantiate this class directly?

beekeeper-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractorTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private static final String EXTRACTOR_KEY = "beekeeper.expired.data.retention.period";
+  private static final String EXTRACTOR_DEFAULT_VALUE = "P7D";
+
+  @Mock private AlterTableEvent listenerEvent;
+  private CleanupDelayExtractor cleanupDelayExtractor;
+
+  @BeforeEach
+  public void setup() {
+    cleanupDelayExtractor = new CleanupDelayExtractor(EXTRACTOR_KEY, EXTRACTOR_DEFAULT_VALUE);
```
massdosage, line 43: You can inline this on line 39 and get rid of this method.

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/service/PathSchedulerApiary.java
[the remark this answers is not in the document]
massdosage: I guess best to leave it for now and just internally think of `SchedulerApiary` as "Scheduler for Apiary".

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingEntity.java
vedantchokshi: True. I've just changed them to getters :-)

beekeeper-core/src/test/java/com/expediagroup/beekeeper/core/repository/HousekeepingMetadataRepositoryTest.java
```diff
@@ -0,0 +1,278 @@
...
+        .housekeepingStatus(SCHEDULED)
+        .creationTimestamp(creationTimestamp)
+        .modifiedTimestamp(creationTimestamp)
+        .cleanupDelay(Duration.parse("P3D"))
+        .cleanupAttempts(0)
+        .lifecycleType(EXPIRED.toString())
+        .build();
+  }
+
+  private void compare(HousekeepingMetadata actual, HousekeepingMetadata expected) {
```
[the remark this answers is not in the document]
vedantchokshi: Sure :-)

beekeeper-scheduler-apiary/src/main/resources/db/migration/V2_1__Create_unpartitioned_table_housekeeping_table_and_alter_path_table_name_and_rename_path_status_column.sql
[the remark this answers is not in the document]
max-jacobs: I think probably each statement should be moved to its own separate file? What's best practice here, do the flyway docs advise this?

beekeeper-path-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/service/PathSchedulerApiary.java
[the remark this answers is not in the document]
max-jacobs: Ah just realised the implications for this sorry. The image will still have "path" in the name, which is fine. Should still run. But we should have a further ticket change the name and then support that in & ecs/eks.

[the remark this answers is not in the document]
max-jacobs: And a couple lines in the Dockerfile will have to change too, otherwise all good

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
...
+### Added
+- Add basic Time To Live (TTL) for all tables.
+- DB migration for creating new `housekeeping_metadata` table.
+### Changed
+- Renamed `beekeeper-path-scheduler-apiary` module to `beekeeper-scheduler-apiary`.
+- Renamed `beekeeper-path-apiary` module to `beekeeper-apiary`.
+- Renamed `beekeeper-assembly-path-scheduler-apiary` module to `beekeeper-assembly-scheduler-apiary`.
+- DB migration to rename `path_status` column to `housekeeping_status` in the `path` table.
+- DB migration to rename `path` table to `housekeeping_path` table.
+
```
patduin, line 17: please add:
```- This is a breaking change and an existing database needs to be migrated please see the README.md#MigratingFrom2.1.0-3.0.0 for instructions.
```
And write up a little section explaining what needs to be done.
 E.g. "run the following script(s) which you can find here:....." etc..

This needs to be a bit more clear our future self will thank us :)

beekeeper-scheduler-apiary/src/main/resources/db/migration/V2_1__Create_unpartitioned_table_housekeeping_table_and_alter_path_table_name_and_rename_path_status_column.sql
```diff
@@ -0,0 +1,19 @@
...
+  creation_timestamp TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  modified_timestamp TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  cleanup_timestamp TIMESTAMP NOT NULL,
+  cleanup_attempts INT NOT NULL DEFAULT 0,
+  client_id VARCHAR(512),
+  lifecycle_type varchar(255) NOT NULL,
+  PRIMARY KEY (id)
+);
+
+ALTER TABLE path RENAME TO housekeeping_path;
```
max-jacobs, line 18: I think that possibly it will be better to have these commands in separate files? What do you think? Just thinking that if the create table succeeds, then another fails, we'll have to roll back the entire thing.

beekeeper-package/beekeeper-assembly-scheduler-apiary/pom.xml
```diff
@@ -5,10 +5,10 @@
   <parent>
     <artifactId>beekeeper-package</artifactId>
     <groupId>com.expediagroup</groupId>
-    <version>2.1.1-SNAPSHOT</version>
+    <version>3.0.0-SNAPSHOT</version>
   </parent>
 
-  <artifactId>beekeeper-assembly-path-scheduler-apiary</artifactId>
+  <artifactId>beekeeper-assembly-scheduler-apiary</artifactId>
```
max-jacobs, line 11: I would check with this change that the assembly still builds the docker image successfully. You should be able to run the image locally without any complaints that the java command can't find the app jar

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/MySqlTestUtils.java
```diff
@@ -61,59 +62,59 @@
     connection = DriverManager.getConnection(jdbcUrl, username, password);
   }
 
-  void insertPath(String path, String table) throws SQLException {
+  void insertPath(String beekeeperTableName, String path, String table) throws SQLException {
```
max-jacobs, line 65: Why did this method change? Looks like its only used in one way

beekeeper-package/pom.xml
```diff
@@ -5,15 +5,15 @@
...
+    <version>3.0.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>beekeeper-package</artifactId>
   <packaging>pom</packaging>
 
   <modules>
     <module>beekeeper-assembly-cleanup</module>
-    <module>beekeeper-assembly-path-scheduler-apiary</module>
+    <module>beekeeper-assembly-scheduler-apiary</module>
```
max-jacobs, line 16: If we're going to go through with this name change, I think we should create some tickets which propagate the name change through all of the terraform and dashboards

beekeeper-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractorTest.java
```diff
@@ -0,0 +1,62 @@
...
+  @Test
+  public void typicalExtractCleanUpDelay() {
+    String cleanupDelayInput = "P30D";
+    when(listenerEvent.getTableParameters()).thenReturn(Map.of(EXTRACTOR_KEY, cleanupDelayInput));
+    Duration cleanupDelay = cleanupDelayExtractor.extractCleanupDelay(listenerEvent);
+    assertThat(cleanupDelay).isEqualTo(Duration.parse(cleanupDelayInput));
+  }
+
+  @Test
+  public void NoKeyExtractCleanUpDelay() {
```
max-jacobs, line 50:
```suggestion
  public void noKeyExtractCleanUpDelay() {
```

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
```diff
@@ -0,0 +1,48 @@
...
+
+  public Duration extractCleanupDelay(ListenerEvent listenerEvent) {
+    String tableCleanupDelay = listenerEvent.getTableParameters().getOrDefault(propertyKey, defaultValue);
+
+    try {
+      return Duration.parse(tableCleanupDelay);
+    } catch (DateTimeParseException e) {
+      log.error("Text '{}' cannot be parsed to a Duration for table '{}.{}'. Using default setting.",
+          tableCleanupDelay, listenerEvent.getDbName(), listenerEvent.getTableName());
+      return Duration.parse(defaultValue);
```
max-jacobs, line 45: I think we should parse this in the constructor so that we don't have to keep on doing it, and so that we can catch any misconfiguration when Beekeeper starts up and not when paths are scheduled.

beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java
```diff
@@ -0,0 +1,83 @@
...
+    return LIFECYCLE_EVENT_TYPE;
+  }
+
+  @Override
+  @TimedTaggable("metadata-scheduled")
+  public void scheduleForHousekeeping(HousekeepingEntity housekeepingEntity) {
+    HousekeepingMetadata housekeepingTable = createOrUpdateHousekeepingTable(
+        (HousekeepingMetadata) housekeepingEntity);
+    try {
+      housekeepingMetadataRepository.save(housekeepingTable);
```
max-jacobs, line 57: maybe it would be good to log something here with some basic information about what's being scheduled. the same goes for the unreferenced handler. currently nothing is logged which is a little annoying.

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
massdosage: I'm generally a fan for failing as fast as possible too.

beekeeper-package/pom.xml
max-jacobs: Heads up that the module's `artifactId` hasn't changed (`beekeeper-assembly-path-scheduler-apiary`). We want to change this too right?

max-jacobs: @patduin?

patduin: yes that should match, I actually would expect maven to complain, good catch.

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
max-jacobs: @vedantchokshi

beekeeper-package/beekeeper-assembly-scheduler-apiary/pom.xml
vedantchokshi: Yup seems to be working

beekeeper-scheduler-apiary/src/main/resources/db/migration/V2_1__Create_unpartitioned_table_housekeeping_table_and_alter_path_table_name_and_rename_path_status_column.sql
vedantchokshi: True but its all part of 1 migration so I don't know know. If we split it out then do you want the alter commands in different files as well?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/MySqlTestUtils.java
vedantchokshi: Because I was getting some issues with the integration tests when I changed the table name of the `path` table to `housekeeping_path`. The `path` was hardcoded in before so o I changed the code to take in the table name as an option

beekeeper-scheduler-apiary/src/main/resources/db/migration/V2_1__Create_unpartitioned_table_housekeeping_table_and_alter_path_table_name_and_rename_path_status_column.sql
max-jacobs: Yeah i can't find any clear docs on flyway, its probably fine

beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java
max-jacobs: Lets leave this for a different PR :)

PR #57, opened 2020-07-03
Feature/add metadata cleanup module

JayGreeeen:
I have added a new module for the metadata cleanup


pom.xml
```diff
@@ -22,6 +22,7 @@
     <module>beekeeper-integration-tests</module>
     <module>beekeeper-package</module>
     <module>beekeeper-vacuum-tool</module>
+    <module>beekeeper-metadata-cleanup</module>
```
max-jacobs, line 25: module should probably go above the integration tests, i think order here affects build order

beekeeper-metadata-cleanup/src/main/resources/beekeeper-metadata-cleanup-application.yml
```diff
@@ -0,0 +1,16 @@
...
+
+spring:
+  jpa:
+    database: default
+    hibernate:
+      ddl-auto: validate
+    properties.hibernate:
+      dialect: org.hibernate.dialect.MySQL8Dialect
+properties:
+  cleanup-page-size: 500
```
max-jacobs, line 14: nice, we'll probably need all of these properties at some point

beekeeper-metadata-cleanup/pom.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+  xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <artifactId>beekeeper-parent</artifactId>
+    <groupId>com.expediagroup</groupId>
+    <version>2.1.1-SNAPSHOT</version>
```
max-jacobs, line 9: might want to check with @vedantchokshi before merging because I think the other PR bumps this to 3.0.0-SNAPSHOT

JayGreeeen: yeah I was going to wait till his stuff was in before I add anything 👌🏻

pom.xml
JayGreeeen: ah okay, good point. I will move this

beekeeper-metadata-cleanup/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+
+  <artifactId>beekeeper-metadata-cleanup</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.expediagroup</groupId>
+      <artifactId>beekeeper-core</artifactId>
+      <version>${project.version}</version>
+      <exclusions>
+        <!-- if this is on classpath then not-null checks in objects persisted via JPA are ignored -->
```
patduin, line 20: eh?

patduin: this is copied from somewhere or not?

JayGreeeen: yeah, copied from the pom for beekeeper-cleanup

patduin: ok so you want to keep just in case you will add non-null checks?

JayGreeeen: yeah

massdosage: Oh, the pain that was involved in figuring this one out, it's all coming back to me now, make it stop!

beekeeper-metadata-cleanup/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+  
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+        <version>${springframework.boot.version}</version>
+        <configuration>
+          <classifier>spring-boot</classifier>
+          <mainClass>com.expediagroup.beekeeper.metadata.cleanup.BeekeeperMetadataCleanup</mainClass>
```
massdosage, line 43: Opinions on whether the base package name should be `com.expediagroup.beekeeper.metadata.cleanup` versus `com.expediagroup.beekeeper.cleanup.metadata`? The latter would make it more like a sub-package of [beekeeper-cleanup](https://github.com/ExpediaGroup/beekeeper/tree/master/beekeeper-cleanup) which may or may not be what you want. I guess it depends whether the focus here is "metadata" or "cleanup".

beekeeper-package/beekeeper-assembly-metadata-cleanup/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    <dependency>
+      <groupId>com.expediagroup</groupId>
+      <artifactId>beekeeper-metadata-cleanup</artifactId>
+      <version>${project.version}</version>
+      <classifier>spring-boot</classifier>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <finalName>beekeeper-metadata-cleanup-${project.version}</finalName>
```
massdosage, line 29: OK, looks like we've gone with `beekeeper-metadata-cleanup*` as the naming throughout so the package name above is fine as it's consistent with that.

beekeeper-metadata-cleanup/pom.xml
patduin: I think focus is more on metadata although the waters are a bit muddy as we'll probably delete data from this module as well yesterday we discussed a shared module that hold the "deletion" logic. 
The features are broadly split into
Unreferencing data: cleaning up paths only (The old cleanup)
Expiring data: cleaning up metadata (tables/partitions) + corresponding paths (the new metadata cleanup)

PR #58, opened 2020-07-14
Add table expiry for partitioned tables

vedantchokshi:
Fixes #

### DIFF
--- a/beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingMetadata.java
+++ b/beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingMetadata.java
@@ -15,6 +15,8 @@
  */
 package com.expediagroup.beekeeper.core.model;
 
+import static java.lang.String.format;
+
 import java.time.Duration;
 import java.time.LocalDateTime;
 
@@ -41,11 +43,17 @@ public class HousekeepingMetadata implements HousekeepingEntity {
   @GeneratedValue(strategy = GenerationType.IDENTITY)
   private Long id;
 
+  @Column(name = "path", nullable = false)
+  private String path;
+
   @Column(name = "database_name", nullable = false)
   private String databaseName;
 
   @Column(name = "table_name", nullable = false)
   private String tableName;
+
+  @Column(name = "partition_name")
+  private String partitionName;
 
   @Column(name = "housekeeping_status", nullable = false)
   @Enumerated(EnumType.STRING)
@@ -77,13 +85,15 @@ public class HousekeepingMetadata implements HousekeepingEntity {
 
   }
 
-  private HousekeepingMetadata(Long id, String databaseName, String tableName, HousekeepingStatus housekeepingStatus,
-      LocalDateTime creationTimestamp, LocalDateTime modifiedTimestamp,
-      LocalDateTime cleanupTimestamp, Duration cleanupDelay, int cleanupAttempts, String lifecycleType,
-      String clientId) {
+  private HousekeepingMetadata(Long id, String path, String databaseName, String tableName,
+      String partitionName, HousekeepingStatus housekeepingStatus, LocalDateTime creationTimestamp,
+      LocalDateTime modifiedTimestamp, LocalDateTime cleanupTimestamp, Duration cleanupDelay, int cleanupAttempts,
+      String lifecycleType, String clientId) {
     this.id = id;
+    this.path = path;
     this.databaseName = databaseName;
     this.tableName = tableName;
+    this.partitionName = partitionName;
     this.housekeepingStatus = housekeepingStatus;
     this.creationTimestamp = creationTimestamp;
     this.modifiedTimestamp = modifiedTimestamp;
@@ -109,6 +119,11 @@ public class HousekeepingMetadata implements HousekeepingEntity {
   }
 
   @Override
+  public String getPath() {
+    return path;
+  }
+
+  @Override
   public String getDatabaseName() {
     return databaseName;
   }
@@ -126,6 +141,10 @@ public class HousekeepingMetadata implements HousekeepingEntity {
     this.tableName = tableName;
   }
 
+  public String getPartitionName() {
+    return partitionName;
+  }
+
   @Override
   public HousekeepingStatus getHousekeepingStatus() {
     return housekeepingStatus;
@@ -140,26 +159,14 @@ public class HousekeepingMetadata implements HousekeepingEntity {
     return creationTimestamp;
   }
 
-  public void setCreationTimestamp(LocalDateTime creationTimestamp) {
-    this.creationTimestamp = creationTimestamp;
-  }
-
   @Override
   public LocalDateTime getModifiedTimestamp() {
     return modifiedTimestamp;
   }
 
-  public void setModifiedTimestamp(LocalDateTime modifiedTimestamp) {
-    this.modifiedTimestamp = modifiedTimestamp;
-  }
-
   @Override
   public LocalDateTime getCleanupTimestamp() {
     return cleanupTimestamp;
-  }
-
-  public void setCleanupTimestamp(LocalDateTime cleanupTimestamp) {
-    this.cleanupTimestamp = cleanupTimestamp;
   }
 
   @Override
@@ -187,7 +194,7 @@ public class HousekeepingMetadata implements HousekeepingEntity {
 
   public void setCleanupDelay(Duration cleanupDelay) {
     this.cleanupDelay = cleanupDelay;
-    cleanupTimestamp = creationTimestamp.plus(cleanupDelay);
+    this.cleanupTimestamp = this.creationTimestamp.plus(cleanupDelay);
   }
 
   @Override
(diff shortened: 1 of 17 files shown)


beekeeper-core/src/test/java/com/expediagroup/beekeeper/core/repository/HousekeepingMetadataRepositoryTest.java
```diff
@@ -57,8 +57,13 @@
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
 public class HousekeepingMetadataRepositoryTest {
```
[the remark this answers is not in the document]
vedantchokshi: This is in the [ExpiredMetadataHousekeepingGeneratorTest](https://github.com/ExpediaGroup/beekeeper/blob/3461c4ad35e5953f00ea9f2f34fe886b72ac447c/beekeeper-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/apiary/generator/ExpiredHousekeepingMetadataGeneratorTest.java) because that's where all the different cases are handled.

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/ExpiredHousekeepingMetadataGenerator.java
```diff
@@ -74,17 +91,46 @@ public LifecycleEventType getLifecycleEventType() {
...
+  private HousekeepingEntity generateHousekeepingEntity(AddPartitionEvent listenerEvent, String clientId) {
+    String partitionName = generatePartitionName(Lists.newArrayList(listenerEvent.getPartitionKeys().keySet()),
+        listenerEvent.getPartitionValues());
+    return generateHousekeepingEntity(listenerEvent, clientId, listenerEvent.getPartitionLocation(), partitionName);
+  }
+
+  private HousekeepingEntity generateHousekeepingEntity(AlterPartitionEvent listenerEvent, String clientId) {
+    String partitionName = generatePartitionName(Lists.newArrayList(listenerEvent.getPartitionKeys().keySet()),
+        listenerEvent.getPartitionValues());
+    return generateHousekeepingEntity(listenerEvent, clientId, listenerEvent.getPartitionLocation(), partitionName);
```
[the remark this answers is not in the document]
vedantchokshi: This is because database and table name are common to all events but the path is unique to the type of ListenerEvent. For example, from `AddPartitionEvent` we need partition location but from `AlterTableEvent` we need table location.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingMetadata.java
```diff
@@ -126,6 +141,10 @@ public void setTableName(String tableName) {
     this.tableName = tableName;
   }
 
+  public String getPartitionName() {
```
patduin, line 144: Can you add a line fo javadoc explaining the format of this name?

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/ExpiredHousekeepingMetadataGenerator.java
```diff
@@ -59,8 +67,17 @@ public ExpiredHousekeepingMetadataGenerator(String cleanupDelay) {
     List<HousekeepingEntity> housekeepingEntities = new ArrayList<>();
 
     switch (listenerEvent.getEventType()) {
+    case CREATE_TABLE:
+      housekeepingEntities.add(generateHousekeepingEntity((CreateTableEvent) listenerEvent, clientId));
+      break;
     case ALTER_TABLE:
-      housekeepingEntities.add(generateHousekeepingEntity(listenerEvent, clientId));
+      housekeepingEntities.add(generateHousekeepingEntity((AlterTableEvent) listenerEvent, clientId));
```
patduin, line 74: this and alter_partition will create a duplicate entry right? You want to add alter because of a location change to capture that correct? But if there is no location change do we need a duplicate entry?

vedantchokshi: There won't be a duplicate entry because we do a look up in the scheduler to find the existing record and update that. Although it will make a new entry if the table or partition name has changed but we need to change around of lot of things to handle that case.

https://github.com/ExpediaGroup/beekeeper/blob/ae0ae70e9a1f3310cb49cb22669d7f8369ff4cad/beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java#L68

patduin: I think a new entry if the table is changes makes perfect sense. We can keep the old entry it will just have nothing to delete as the table is gone already.
So this caputures alters on the cleanupDelay property only right?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingMetadata.java
patduin: @vedantchokshi ^

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/ExpiredHousekeepingMetadataGenerator.java
vedantchokshi: Yeah exactly

patduin: ok let's go with this for now

beekeeper-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/apiary/generator/HousekeepingEntityGeneratorTestBase.java
```diff
@@ -61,11 +62,13 @@ protected void setupClockAndExtractor(ListenerEvent listenerEvent) {
 
   protected void assertHousekeepingEntity(HousekeepingEntity housekeepingEntity,
       LifecycleEventType lifecycleEventType) {
+    System.out.println(housekeepingEntity);
```
patduin, line 65: remove

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingMetadata.java
patduin: ???

beekeeper-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/apiary/generator/HousekeepingEntityGeneratorTestBase.java
vedantchokshi, line 65:
```suggestion
    System.out.println(housekeepingEntity);
```

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingMetadata.java
massdosage: Looks good, also matches how Hive seem to describe this in the (very limited) javadoc on https://hive.apache.org/javadocs/r2.1.1/api/org/apache/hadoop/hive/metastore/IMetaStoreClient.html#partitionNameToVals-java.lang.String-

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingMetadata.java
```diff
@@ -187,19 +201,29 @@ public Duration getCleanupDelay() {
 
   public void setCleanupDelay(Duration cleanupDelay) {
     this.cleanupDelay = cleanupDelay;
-    cleanupTimestamp = creationTimestamp.plus(cleanupDelay);
+    this.cleanupTimestamp = this.creationTimestamp.plus(cleanupDelay);
```
massdosage, line 204: Minor, but you don't need the `this` here as it's not ambiguous what you're referring to unlike `cleanupDelay` which in this case is both a method parameter and a member variable.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingMetadata.java
```diff
@@ -187,19 +201,29 @@ public Duration getCleanupDelay() {
...
   }
 
   @Override
   public MetricTag getMetricTag() {
     return new MetricTag("table", String.join(".", databaseName, tableName));
   }
 
+  @Override
+  public String toString() {
+    return format(
```
massdosage, line 214: The last two params are the wrong way around.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingPath.java
```diff
@@ -198,14 +192,22 @@ public Duration getCleanupDelay() {
 
   public void setCleanupDelay(Duration cleanupDelay) {
     this.cleanupDelay = cleanupDelay;
-    cleanupTimestamp = creationTimestamp.plus(cleanupDelay);
+    this.cleanupTimestamp = this.creationTimestamp.plus(cleanupDelay);
```
massdosage, line 195: As above.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/HousekeepingPath.java
```diff
@@ -198,14 +192,22 @@ public Duration getCleanupDelay() {
...
   }
 
   @Override
   public MetricTag getMetricTag() {
     return new MetricTag("table", String.join(".", databaseName, tableName));
   }
 
+  @Override
+  public String toString() {
```
massdosage, line 204: Same problem here.

PR #61, opened 2020-07-27
Feature/cleanup expired tables

JayGreeeen:
PR for cleaning up expired metadata. 
I have moved the code for dropping tables and deleting paths into core, as both the cleanup and metadata-cleanup modules will delete paths. 

Contains commits from adding the metadata cleanup module and changes made in Vedants PR.


beekeeper-core/pom.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
vedantchokshi, line 1: I'm not sure if we should move all the common path-cleaner stuff to the core module. The core module is used by scheduler and everything else which means the clean up stuff will be available to pretty much every other beekeeper module we make. 

I think it might be a good idea to follow the same pattern we have for the scheduler. So `beekeeper-scheduler` is the generic module for scheduling an entity and then `beekeeper-scheduler-apiary` uses that module to schedule for apiary events. Similarly, we could have the `beekeeper-cleanup` module with the common clean up stuff and then `beekeeper-cleanup-path` and `beekeeper-cleanup-metadata` make use of the clean up module. What do you think?

@massdosage @patduin @abhimanyugupta07

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/config/MetadataType.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.expediagroup.beekeeper.core.config;
+
+public enum MetadataType {
+
```
vedantchokshi, line 19: Are the blank lines supposed to be there?

beekeeper-core/pom.xml
JayGreeeen: Yeah thats a good point, and would make sense to do it in the same way as the scheduler stuff.

abhimanyugupta07: Sounds good !

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/hive/HiveMetadataCleaner.java
```diff
@@ -0,0 +1,52 @@
...
+  private HiveClient client;
+  private DeletedMetadataReporter deletedMetadataReporter;
+
+  public HiveMetadataCleaner(HiveClient client, DeletedMetadataReporter deletedMetadataReporter) {
+    this.client = client;
+    this.deletedMetadataReporter = deletedMetadataReporter;
+  }
+
+  @Override
+  public void cleanupMetadata(HousekeepingMetadata housekeepingMetadata) {
```
vedantchokshi, line 34: Could we maybe use `cleanupTableMetadata` and `cleanupPartitionMetadata` because they're both metadata?

beekeeper-core/src/test/java/com/expediagroup/beekeeper/core/hive/HiveClientTest.java
```diff
@@ -0,0 +1,131 @@
...
+import org.mockito.Mock;
+import org.mockito.Mockito;
+
+import com.expediagroup.beekeeper.core.error.BeekeeperException;
+
+import com.hotels.hcommon.hive.metastore.client.api.CloseableMetaStoreClient;
+
+public class HiveClientTest {
+
+  private @Mock CloseableMetaStoreClient client = Mockito.mock(CloseableMetaStoreClient.class);
```
vedantchokshi, line 40: I don't think you need to create the mock object after you use the mock annotation
```suggestion
  @Mock 
  private CloseableMetaStoreClient client;
```

beekeeper-core/src/test/java/com/expediagroup/beekeeper/core/hive/HiveClientTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    boolean result = hiveClient.dropPartition(DATABASE, TABLE_NAME, PARTITION_NAME);
+    verify(client).dropPartition(DATABASE, TABLE_NAME, PARTITION_NAME, false);
+    assertFalse(result);
+  }
+
+  @Test
+  public void throwsExceptionForTableDelete() throws MetaException, NoSuchObjectException, TException {
+    Mockito.doThrow(MetaException.class).when(client).dropTable(DATABASE, TABLE_NAME);
+    Assertions.assertThrows(BeekeeperException.class, () -> {
```
vedantchokshi, line 117: Static import for `assertThrows`?

beekeeper-core/pom.xml
massdosage: +1

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/hive/HiveClient.java
```diff
@@ -0,0 +1,111 @@
...
+    this.dryRunEnabled = dryRunEnabled;
+  }
+
+  /**
+   * Will delete the table metadata. Error is not thrown if table not found.
+   * 
+   * @param database
+   * @param tableName
+   */
+  public boolean deleteMetadata(String database, String tableName) {
```
massdosage, line 45: Wouldn't `dropTable` be a more accurate method name? Or have you gone with this name to distinguish between deleting the data in the table? I still think I'd prefer the `dropTable` name as the behaviour is more consistent with how Hive works and this is a `HiveClient` class.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/hive/HiveClient.java
```diff
@@ -0,0 +1,111 @@
...
+      }
+    }
+    return partitionDeleted;
+  }
+
+  private boolean tableExists(String databaseName, String tableName) {
+    try {
+      return client.tableExists(databaseName, tableName);
+    } catch (TException e) {
+      return false;
```
massdosage, line 107: Do you not want to rethrow this? Surely if this happens something has gone wrong and maybe it's not safe to assume that the table doesn't exist?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/hive/HiveMetadataCleaner.java
massdosage: +1

beekeeper-core/src/test/java/com/expediagroup/beekeeper/core/repository/HousekeepingMetadataRepositoryTest.java
```diff
@@ -252,12 +252,85 @@ public void findRecordForCleanupByDatabaseAndTableMixedHousekeepingStatus() {
...
+    HousekeepingMetadata housekeepingTable1 = createPartitionedEntityHousekeepingTable();
+    housekeepingMetadataRepository.save(housekeepingTable1);
+    HousekeepingMetadata housekeepingTable2 = createEntityHouseKeepingTable(DATABASE_NAME + "1", TABLE_NAME + "1",
+        PARTITION_NAME);
+    housekeepingMetadataRepository.save(housekeepingTable2);
+
+    Page<HousekeepingMetadata> result = housekeepingMetadataRepository
+        .findRecordsForGivenDatabaseAndTable(DATABASE_NAME, TABLE_NAME, PageRequest.of(0, 500));
+
+    assertTrue(result.getContent().size() == 1);
```
massdosage, line 294: Rather use `assertEquals` or `assertThat....` as one gets a better failure message if this isn't true.

beekeeper-core/src/test/java/com/expediagroup/beekeeper/core/repository/HousekeepingMetadataRepositoryTest.java
```diff
@@ -252,12 +252,85 @@ public void findRecordForCleanupByDatabaseAndTableMixedHousekeepingStatus() {
...
+    HousekeepingMetadata housekeepingTable1 = createPartitionedEntityHousekeepingTable();
+    housekeepingTable1.setHousekeepingStatus(DELETED);
+    housekeepingMetadataRepository.save(housekeepingTable1);
+    HousekeepingMetadata housekeepingTable2 = createUnpartitionedEntityHousekeepingTable();
+    housekeepingMetadataRepository.save(housekeepingTable2);
+
+    Page<HousekeepingMetadata> result = housekeepingMetadataRepository
+        .findRecordsForGivenDatabaseAndTable(DATABASE_NAME, TABLE_NAME, PageRequest.of(0, 500));
+
+    assertTrue(result.getContent().size() == 1);
```
massdosage, line 309: ditto

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandler.java
```diff
@@ -0,0 +1,165 @@
...
+  public abstract PathCleaner getPathCleaner();
+
+  public abstract Page<HousekeepingMetadata> findRecordsToClean(LocalDateTime instant, Pageable pageable);
+
+  public abstract Page<HousekeepingMetadata> findMatchingRecords(
+      String databaseName,
+      String tableName,
+      Pageable pageable);
+
+  /*-
```
massdosage, line 52: What does this body of comments apply to? One of the methods below or in general just to how this class behaves?

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandler.java
```diff
@@ -0,0 +1,165 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  private void cleanUpTable(
+      HousekeepingMetadata housekeepingMetadata,
+      MetadataCleaner metadataCleaner,
+      PathCleaner pathCleaner) {
+    metadataCleaner.cleanupMetadata(housekeepingMetadata);
```
massdosage, line 116: OK, seeing how it's used I think it's fine to rename `cleanupMetadata` to `dropTable`? Makes it clearer what's going on?

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandler.java
```diff
@@ -0,0 +1,165 @@
...
+    pathCleaner.cleanupPath(housekeepingMetadata);
+  }
+
+  private void cleanupPartition(
+      HousekeepingMetadata housekeepingMetadata,
+      MetadataCleaner metadataCleaner,
+      PathCleaner pathCleaner) {
+    metadataCleaner.cleanupPartition(housekeepingMetadata);
```
massdosage, line 124:
```suggestion
    metadataCleaner.dropPartition(housekeepingMetadata);
```
? Unless there is some reason for using `cleanup` that I've missed?

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandler.java
```diff
@@ -0,0 +1,165 @@
...
+      return true;
+    }
+    if (partitionName != null) {
+      cleanupPartition(housekeepingMetadata, metadataCleaner, pathCleaner);
+      return true;
+    }
+    return false;
+  }
+
+  private void cleanUpTable(
```
vedantchokshi, line 112: Given our new decision to use unreferenced, I think we should only delete the table or partition if it exists in hive **and** also the data exists. This is because say you alter the partition name then it will create a new entry the in housekeeping table. Because the old partition name will still exist in the table, this process will try to drop the partition but won't find it but it will still drop the data and it could lead to confusion.

JayGreeeen: Do you mean only drop the data if the table/ partition exists? i.e. drop the table/partition, if it doesnt exist then dont delete the data - as itll be picked up by unreferenced.

beekeeper-core/src/test/java/com/expediagroup/beekeeper/core/hive/HiveClientTest.java
JayGreeeen: It actually doesnt work if you dont create the mock object, but I can remove the annotation 👍

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandler.java
JayGreeeen: just generally how the class behaves, I added it when starting the cleanup module as a helpful explanation for how the paging will work - its slightly different to the path cleanup. I can put it somewhere else or just take it out entirely if we think it's not needed.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3DryRunPathCleanerTest.java
```diff
@@ -54,6 +54,7 @@
   private AmazonS3 amazonS3;
   private S3Client s3Client;
   private @Mock BytesDeletedReporter bytesDeletedReporter;
+  private Boolean dryRunEnabled = true;
```
massdosage, line 57: Minor, but this could be a `boolean`.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3PathCleanerTest.java
```diff
@@ -75,6 +75,7 @@
   private S3Client s3Client;
   private S3SentinelFilesCleaner s3SentinelFilesCleaner;
   private @Mock BytesDeletedReporter bytesDeletedReporter;
+  private Boolean dryRunEnabled = false;
```
massdosage, line 78: Minor, but this could be a `boolean`.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/hive/HiveClientTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import com.expediagroup.beekeeper.core.error.BeekeeperException;
+
+import com.hotels.hcommon.hive.metastore.client.api.CloseableMetaStoreClient;
+
+public class HiveClientTest {
+
+  private CloseableMetaStoreClient client = Mockito.mock(CloseableMetaStoreClient.class);
```
massdosage, line 39:
```suggestion
  private @Mock CloseableMetaStoreClient client;
```

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporterTest.java
```diff
@@ -53,10 +53,9 @@
   private static final String TABLE = "database.table";
   private static final long BYTES_DELETED = 10;
 
-  @Autowired
-  private MeterRegistry meterRegistry;
-  @Mock
-  private Taggable taggable;
+  private @Autowired MeterRegistry meterRegistry;
+  private @Mock
+  Taggable taggable;
```
massdosage, line 58:
```suggestion
  private @Mock Taggable taggable;
```

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/DeletedMetadataReporterTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private static final String TABLE = "database.table";
+
+  private String tableCounterName = String.join("-", MetadataType.HIVE_TABLE.toString().toLowerCase(), METRIC_NAME);
+  private String dryRunTableCounterName = String
+      .join("-", MetadataType.HIVE_TABLE.toString().toLowerCase(), DRY_RUN_METRIC_NAME);
+
+  private @Autowired MeterRegistry meterRegistry;
+  private @Mock
+  Taggable taggable;
```
massdosage, line 60:
```suggestion
  private @Mock Taggable taggable;
```

beekeeper-metadata-cleanup/src/test/java/com/expediagroup/beekeeper/metadata/cleanup/context/CommonBeansTest.java
```diff
@@ -0,0 +1,161 @@
...
+  private static final String AWS_S3_ENDPOINT_PROPERTY = "aws.s3.endpoint";
+  private static final String AWS_REGION_PROPERTY = "aws.region";
+  private static final String REGION = "us-west-2";
+  private static final String AWS_ENDPOINT = String.join(".", "s3", REGION, "amazonaws.com");
+  private static final String ENDPOINT = "endpoint";
+  private static final String BUCKET = "bucket";
+  private static final String KEY = "key";
+  private final String metastoreUri = "thrift://localhost:1234";
+
+  private Boolean dryRunEnabled = false;
```
massdosage, line 67:
```suggestion
  private boolean dryRunEnabled = false;
```

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/hive/HiveMetadataCleanerTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import com.expediagroup.beekeeper.cleanup.monitoring.DeletedMetadataReporter;
+import com.expediagroup.beekeeper.core.config.MetadataType;
+import com.expediagroup.beekeeper.core.model.HousekeepingMetadata;
+
+@ExtendWith(MockitoExtension.class)
+public class HiveMetadataCleanerTest {
+
+  private @Mock HousekeepingMetadata housekeepingMetadata;
```
vedantchokshi, line 35: Not sure about this suggestion but I don't think there's any need to use a mock `HousekeepingMetadata` instance here when we can just use an actual one. You can remove the verify calls because the `verify(hiveClient).dropTable(DATABASE, TABLE_NAME)` indirectly verifies that the right methods are called.  You can also spy the actual object if you want and the verify the method calls.

I've always been under the impression to only use mocks when it is too difficult to create the object. @patduin care to share some of your wisdom with us? :-)

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/hive/HiveMetadataCleanerTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @BeforeEach
+  public void init() {
+    cleaner = new HiveMetadataCleaner(hiveClient, deletedMetadataReporter);
+    when(housekeepingMetadata.getDatabaseName()).thenReturn(DATABASE);
+    when(housekeepingMetadata.getTableName()).thenReturn(TABLE_NAME);
+  }
+
+  @Test
+  public void typicalMetadataCleanup() {
```
vedantchokshi, line 52: Can we call this `typicalTableCleanup()` so its more clear?

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/hive/HiveClientTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+  private HiveClient hiveClient;
+
+  @BeforeEach
+  public void init() {
+    hiveClient = new HiveClient(client, false);
+  }
+
+  @Test
+  public void typicalMetadataDelete() throws MetaException, NoSuchObjectException, TException {
```
vedantchokshi, line 57: `typicalTableDelete()`?

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/hive/HiveClientTest.java
```diff
@@ -0,0 +1,133 @@
...
+  @Test
+  public void typicalPartitionDrop() throws NoSuchObjectException, MetaException, TException {
+    when(client.tableExists(DATABASE, TABLE_NAME)).thenReturn(true);
+    boolean result = hiveClient.dropPartition(DATABASE, TABLE_NAME, PARTITION_NAME);
+    verify(client).dropPartition(DATABASE, TABLE_NAME, PARTITION_NAME, false);
+    assertTrue(result);
+  }
+
+  @Test
+  public void typicalMetadataDryRun() throws MetaException, NoSuchObjectException, TException {
```
vedantchokshi, line 72: `typicalTableDryRun`

beekeeper-metadata-cleanup/src/test/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandlerTest.java
```diff
@@ -0,0 +1,242 @@
...
+    verify(pathCleaner, never()).cleanupPath(mockMetadata);
+    verify(metadataCleaner, never()).dropTable(mockMetadata);
+    verify(mockMetadata).setCleanupAttempts(1);
+    verify(mockMetadata).setHousekeepingStatus(DELETED);
+    verify(housekeepingMetadataRepository).save(mockMetadata);
+    assertThat(pageable).isEqualTo(pageable);
+  }
+
+  @Test
+  public void expectedMetadataCleanupFailure() {
```
vedantchokshi, line 198: `expectedTableCleanupFailure`?

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandler.java
```diff
@@ -0,0 +1,146 @@
...
+import org.springframework.data.domain.Pageable;
+
+import com.expediagroup.beekeeper.cleanup.metadata.MetadataCleaner;
+import com.expediagroup.beekeeper.cleanup.path.PathCleaner;
+import com.expediagroup.beekeeper.core.model.HousekeepingMetadata;
+import com.expediagroup.beekeeper.core.model.HousekeepingStatus;
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+import com.expediagroup.beekeeper.core.repository.HousekeepingMetadataRepository;
+
+public abstract class GenericMetadataHandler {
```
patduin, line 33: Add an interface MetadataHandler with just the two methods that need to be exposed:
 - findRecordsToClean
- processPage

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandler.java
```diff
@@ -0,0 +1,146 @@
...
+
+  private final Logger log = LoggerFactory.getLogger(GenericMetadataHandler.class);
+
+  public abstract HousekeepingMetadataRepository getHousekeepingMetadataRepository();
```
patduin, line 37: none of these methods (line 37 to line 43) should be public, can we just have a constructor that takes all the building blocks it needs? Then all these methods can just go.

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandler.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public abstract LifecycleEventType getLifecycleType();
+
+  public abstract MetadataCleaner getMetadataCleaner();
+
+  public abstract PathCleaner getPathCleaner();
+
+  public abstract Page<HousekeepingMetadata> findRecordsToClean(LocalDateTime instant, Pageable pageable);
```
patduin, line 45: this method should go in the interface

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandler.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public abstract Long countPartitionsForDatabaseAndTable(LocalDateTime instant, String databaseName, String tableName, boolean dryRunEnabled);
```
patduin, line 47: this should be protected

beekeeper-metadata-cleanup/src/test/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandlerTest.java
```diff
@@ -0,0 +1,255 @@
...
+  private static final LocalDateTime CLEANUP_INSTANCE = LocalDateTime.ofInstant(Instant.now(), ZoneOffset.UTC);
+
+  @BeforeEach
+  public void init() {
+    when(mockPage.getContent()).thenReturn(List.of(mockMetadata));
+    when(mockMetadata.getDatabaseName()).thenReturn(DATABASE);
+    when(mockMetadata.getTableName()).thenReturn(TABLE_NAME);
+    when(mockMetadata.getPartitionName()).thenReturn(null);
+
+    handler = new ExpiredMetadataHandler(housekeepingMetadataRepository, metadataCleaner, pathCleaner);
```
patduin, line 72: So you're actually testing an implementation. I would remove this class and move all the tests to ExpiredMetadataHandlerTest as that is the only implementation you're actually need to test.

beekeeper-metadata-cleanup/src/test/java/com/expediagroup/beekeeper/metadata/cleanup/service/PagingMetadataCleanupServiceTest.java
```diff
@@ -0,0 +1,285 @@
...
+            createHousekeepingMetadata("table3", "s3://some_foobar", null, FAILED));
+
+    for (int i = 0; i < 5; i++) {
+      int finalI = i;
+      tables.forEach(path -> {
+        if (finalI == 0) {
+          metadataRepository.save(path);
+        }
+
+        doThrow(new RuntimeException("Error")).when(metadataCleaner).dropTable(Mockito.any());
```
patduin, line 235: you don't need to mock this all the time I think in the loop, you can just set it outside the for loop

beekeeper-metadata-cleanup/src/test/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandlerTest.java
```diff
@@ -0,0 +1,255 @@
...
+    when(metadataCleaner.tableExists(DATABASE, TABLE_NAME)).thenReturn(true);
+
+    Pageable pageable = handler.processPage(mockPageable, CLEANUP_INSTANCE, mockPage, false);
+    verify(metadataCleaner).dropTable(mockMetadata);
+    verify(pathCleaner).cleanupPath(mockMetadata);
+    verify(metadataCleaner, never()).dropPartition(mockMetadata);
+    verify(mockMetadata).setCleanupAttempts(1);
+    verify(mockMetadata).setHousekeepingStatus(DELETED);
+    verify(housekeepingMetadataRepository).save(mockMetadata);
+    assertThat(pageable).isEqualTo(pageable);
```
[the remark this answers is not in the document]
JayGreeeen: Its making sure that `pageable.next()` hasnt been called. That should only happen in a dry run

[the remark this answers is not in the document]
patduin: ah perhaps:
```assertThat(pageable).isEqualTo(mockPageable);```
or
```verify(mockPageable, never()).next();```
is more clear.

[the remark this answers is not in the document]
Comment (JayGreeeen) [reply beekeeper-metadata-cleanup/src/test/java/com/expediagroup/beekeeper/metadata/cleanup/handler/GenericMetadataHandlerTest.java:90]: yeah I think the verify is more clear, will change 👍

File: beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveClient.java [hunk 32]
```diff
@@ -0,0 +1,103 @@
...
+    this.dryRunEnabled = dryRunEnabled;
+  }
+
+  /**
+   * Will drop the table from the database if it exists.
+   *
+   * @param databaseName
+   * @param tableName
+   */
+  public void dropTableIfExists(String databaseName, String tableName) {
```
massdosage, line 45: Minor, but you could rename this to just `dropTable` since `dropPartition` below also only does the drop if it exists and doesn't throw an error if not but the method isn't called `dropPartitionIfExists`.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveMetadataCleaner.java
```diff
@@ -0,0 +1,58 @@
...
+    this.client = client;
+    this.deletedMetadataReporter = deletedMetadataReporter;
+  }
+
+  @Override
+  @TimedTaggable("hive-table-deleted")
+  public void dropTable(HousekeepingMetadata housekeepingMetadata) {
+    client
+        .dropTableIfExists(housekeepingMetadata.getDatabaseName(), housekeepingMetadata.getTableName());
+    deletedMetadataReporter.reportTaggable(housekeepingMetadata, MetadataType.HIVE_TABLE);
```
massdosage, line 39: Do you want to have the above return a boolean and only do the increment if it returned true like the code below? Would be more consistent...

massdosage: Do we want a try/catch around this to also catch errors? Or are we OK with them just being thrown and then I assume the only way to know about them is to look in the log? Or we do track them elsewhere?

beekeeper-core/src/test/java/com/expediagroup/beekeeper/core/repository/HousekeepingMetadataRepositoryTest.java
```diff
@@ -252,12 +253,115 @@ public void findRecordForCleanupByDatabaseAndTableMixedHousekeepingStatus() {
...
+    HousekeepingMetadata housekeepingTable1 = createPartitionedEntityHousekeepingTable();
+    housekeepingMetadataRepository.save(housekeepingTable1);
+
+    HousekeepingMetadata housekeepingTable2 = createEntityHouseKeepingTable(DATABASE_NAME, TABLE_NAME + "2", null);
+    housekeepingMetadataRepository.save(housekeepingTable2);
+
+    long result = housekeepingMetadataRepository
+        .countRecordsForGivenDatabaseAndTableWherePartitionIsNotNull(DATABASE_NAME, TABLE_NAME);
+
+    assertEquals(Long.valueOf(1), result);
```
massdosage, line 267:
```suggestion
    assertEquals(1L, result);
```
Can you not just do that? ^

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveMetadataCleaner.java
JayGreeeen: Thats what it was initially, both the `dropTable` and `dropPartition` were the same and would return true if the drop was successful, which would determine whether it should be reported it. This also got returned to the handler class that calls these methods, as the paths are only deleted if the drop is successful.

I spoke to Patrick and he requested I change it to not return true and actually only call this method from the handler if the table exists. But this doesnt entirely work for the partition b/c you can't check it exists before trying to drop it, so thats why its still returning boolean. 

Theres a try/catch in the `dropTableIfExists` method which will catch+log/ throw depending on the error. If its `NoSuchObj` thats fine so its just logged, anything else will be thrown.

massdosage: OK. Are we OK with just logging the errors and not incrementing a counter that we could potentially put a dashboard and alerts on etc.

JayGreeeen: We're expecting that the only time we'll get those errors is when a table is renamed which creates a new entry in the HousekeepingMetadataRepository, or if a user performs the drop themselves, so it shouldnt be something we need to alert on.

massdosage: OK, let's hope you're correct in your expectations ;)

patduin: Having alerts on thrown exceptions is a good idea, we do that for the other Beekeeper modules not sure how this is handled now but if it isn't maybe we create a ticket to tackle this as a next step??

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveClient.java
patduin: yeah I'm fine changing it back as well I asked for the rename to ...IfExists but now we got the trycatch(NoSuchObject..) it makes more sense to just call it dropTable again, sorry Jay.

JayGreeeen: Okay I'll change the name 👍 

Do you also want me to change it back so it returns a boolean like the `dropPartition` method? Or fine as is?

patduin: nah we don't need that and I still don't like that :)

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveMetadataCleaner.java
JayGreeeen: I was talking about the `NoSuchObj` exceptions, these arent thrown, theyre only logged so I'm not sure its worth counting them. But might be worth counting/alerting on the exceptions that do get thrown. 

I'm not sure how its handled but I see theres a graph for it on the hcom dashboard. It isnt in the EGDL dashboard though. We can add a ticket for this to be added in.

patduin: I think noSuchObj doesn't need to be counted. General exceptions should similary to what we do with path deletion.

massdosage (approved):

vedantchokshi (approved):

beekeeper-cleanup/pom.xml
```diff
@@ -12,84 +13,65 @@
...
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-autoconfigure</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter-web</artifactId>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-metastore</artifactId>
+      <version>${hive.version}</version>
+      <scope>provided</scope>
```
patduin, line 40: what's providing this?

beekeeper-metadata-cleanup/pom.xml
```diff
@@ -11,27 +11,93 @@
...
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-metastore</artifactId>
+      <version>${hive.version}</version>
+      <scope>provided</scope>
```
patduin, line 60: same here

beekeeper-cleanup/pom.xml
JayGreeeen, line 40:
```suggestion
```

beekeeper-metadata-cleanup/pom.xml
JayGreeeen, line 60:
```suggestion
```

patduin (approved):

PR #62, opened 2020-08-17
Feature/remove abstract cleanup

JayGreeeen:
Fixes #


beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/MetadataHandler.java
```diff
@@ -16,16 +16,47 @@
...
-      boolean dryRunEnabled);
+  public MetadataHandler(MetadataCleanup metadataCleanup) {
+    this.metadataCleanup = metadataCleanup;
+  }
+
+  /**
+   * Processes a pageable HouseKeepingMetadata page.
+   *
+   * @param pageable Pageable to iterate through for dryRun
+   * @param page Page to get content from
```
massdosage, line 39: `instant` is missing.

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/MetadataHandler.java
```diff
@@ -16,16 +16,47 @@
...
+   * @param dryRunEnabled Dry Run boolean flag
+   * @return Pageable to pass to query. In the case of dry runs, this is the next page.
+   * @implNote This parent handler expects the child's cleanupMetadata call to update & remove the record from this call
```
massdosage, line 42: After this refactor is this still accurate?

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/cleaner/MetadataCleanup.java
```diff
@@ -0,0 +1,32 @@
...
+package com.expediagroup.beekeeper.metadata.cleanup.cleaner;
+
+import java.time.LocalDateTime;
+
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+
+import com.expediagroup.beekeeper.core.model.HousekeepingMetadata;
+
+public interface MetadataCleanup {
```
massdosage, line 25:
```suggestion
public interface MetadataCleaner {
```
thoughts on this rename?

JayGreeeen: We're not sure on the naming we want to go for. Initially Vedant put that, but theres already a `MetadataCleaner` interface inside the beekeeper-cleanup module, which is implemented by the `HiveMetadataCleaner`, so went for this instead

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/context/CommonBeans.java
```diff
@@ -136,9 +139,23 @@ PathCleaner pathCleaner(
...
+  @Bean(name = "expiredMetadataCleanupHandler")
+  public MetadataHandler expiredMetadataCleanupHandler(
+      @Qualifier("expiredMetadataCleanup") ExpiredMetadataCleanup expiredMetadataCleanup) {
+    return new MetadataHandler(expiredMetadataCleanup);
+  }
+
   @Bean
   CleanupService cleanupService(
-      List<GenericMetadataHandler> metadataHandlers,
+          List<MetadataHandler> metadataHandlers,
```
vedantchokshi, line 158: Is this indent meant to be there?

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/cleaner/ExpiredMetadataCleanup.java
```diff
@@ -27,70 +31,64 @@
...
+      updateAttemptsAndStatus(housekeepingMetadata, FAILED);
+      log.warn("Unexpected exception when deleting metadata for table \"{}.{}\"",
+          housekeepingMetadata.getDatabaseName(),
+          housekeepingMetadata.getTableName(), e);
     }
   }
 
-  private boolean cleanupMetadata(HousekeepingMetadata housekeepingMetadata, LocalDateTime instant, boolean dryRunEnabled) {
+  @Override
+  public boolean cleanupMetadata(HousekeepingMetadata housekeepingMetadata, LocalDateTime instant,
```
patduin, line 74: private?

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/cleaner/MetadataCleanup.java
```diff
@@ -0,0 +1,32 @@
...
+
+  Page<HousekeepingMetadata> findRecordsToClean(LocalDateTime instant, Pageable pageable);
+
+  void cleanupContent(HousekeepingMetadata housekeepingMetadata, LocalDateTime instant, boolean dryRunEnabled);
+
+  boolean cleanupMetadata(HousekeepingMetadata housekeepingMetadata, LocalDateTime instant, boolean dryRunEnabled);
```
patduin, line 31: why you need this? it's exactly the same as cleanupContent except for the method name?

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/MetadataHandler.java
```diff
@@ -16,16 +16,48 @@
...
+   * such that subsequent DB queries will not return the record. Hence why we only call next during dryRuns
+   * where no updates occur.
+   * @implNote Note that we only expect pageable.next to be called during a dry run.
+   */
+  public Pageable processPage(Pageable pageable, LocalDateTime instant, Page<HousekeepingMetadata> page,
+      boolean dryRunEnabled) {
+    List<HousekeepingMetadata> pageContent = page.getContent();
+    if (dryRunEnabled) {
+      pageContent.forEach(metadata -> metadataCleanup.cleanupMetadata(metadata, instant, true));
```
patduin, line 52: this is weird with the two methods, also instead of true just pass along the dryRunEnable flag.

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/MetadataHandler.java
```diff
@@ -16,16 +16,48 @@
...
+      return pageable.next();
+    } else {
+      pageContent.forEach(metadata -> metadataCleanup.cleanupContent(metadata, instant, false));
+      return pageable;
+    }
+  }
+
+  public Page<HousekeepingMetadata> findRecordsToClean(LocalDateTime instant, Pageable pageable) {
```
patduin, line 60: Do you need this method, it just delegates doesn't add any functionality

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/cleaner/MetadataCleanup.java
JayGreeeen: https://github.com/ExpediaGroup/beekeeper/blob/91fbd50447/beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/MetadataHandler.java#L51

Its for when theres a dry run. The idea is that `cleanupContent` will call `cleanupMetadata` and then `updateAttemptsAndStatus` - we dont want attempts and status to be updated in a dry run so only cleanup metadata is called instead. 
Since refactoring though its quite unclear, so I'll change this.

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/MetadataHandler.java
JayGreeeen: Yes because this method is used in the `PagingMetadataCleanupService` and it only knows about the `MetadataHandler` and not the `MetadataCleanup`.
https://github.com/ExpediaGroup/beekeeper/blob/91fbd50447/beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/service/PagingMetadataCleanupService.java#L71

patduin: Guess I'm saying if it doesn't add any functionality this class is not doing a whole lot, what functionality does it really add. It certainly doesn't handle metadata perhaps you should just remove it or rename it to something like PageProcessor as that's what this class seems to do??

JayGreeeen: I see your point.. I'm taking it out! 👌

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/cleaner/MetadataCleanup.java
JayGreeeen: I've removed the class that was previously `MetadataHandler` as it wasnt doing a whole lot (just processing pages), so now this interface and its impl have been changed to `MetadataHandler` and `ExpiredMetadataHandler` 👍

PR #63, opened 2020-08-21
Feature/expired tables integration tests

JayGreeeen:
Integration tests for new Metadata Cleanup module.
Contains all the commits from both mine and Vedants previous PRs - I was merging them in as we progressed.


patduin (approved):

beekeeper-integration-tests/pom.xml
```diff
@@ -10,12 +10,40 @@
 
   <artifactId>beekeeper-integration-tests</artifactId>
 
+  <properties>
+    <beeju.version>3.1.0</beeju.version>
+    <hadoop.version>3.3.0</hadoop.version>
```
nvitucci, line 15: Are you actually using Hadoop 3?

JayGreeeen: Using it to important some Constants `org.apache.hadoop.fs.s3a.Constants.ACCESS_KEY`, `org.apache.hadoop.fs.s3a.Constants.ENDPOINT`, `org.apache.hadoop.fs.s3a.Constants.SECRET_KEY`.

massdosage: My guess is that those were only added as constants in Hadoop 3 even though they were available for use in Hadoop 2. I think for now, if you can't find them anywhere in Hadoop 2, declare them as constant strings in our code with a note to replace them with the Hadoop 3 versions (mention the fully qualified class names in a comment) when we upgrade at a later date. This is probably safer than pulling in a Hadoop 3 dependency when all our other stuff using Hadoop 2, there might be some unexpected surprises later down the road which we should avoid.

nvitucci: I agree. This looks like the only place where Hadoop 3 is allegedly used, and this could be risky given that we can use constants instead.

nvitucci (approved):

patduin (approved):

PR #64, opened 2020-09-04
Fixing deployment

JayGreeeen:
Fixing some bits for the deployment. 

Excuse the million commits - I branched off a branch that was branched off a branch 🙃 🤨 🌲


beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java
```diff
@@ -79,6 +83,61 @@ private HousekeepingMetadata createOrUpdateHousekeepingMetadata(HousekeepingMeta
...
+  /**
+   * When the cleanup delay of a table with partitions is altered, the delay should be changed but the cleanup timestamp
+   * should be the max timestamp of any of the partitions which the table has.
+   *
+   * i.e. if the cleanup delay was 10 but now its being updated to 2, the delay should match any partition which has
+   * delay 10 (or above) to prevent premature attempts to cleanup the table.
+   *
+   * @param housekeepingMetadata
+   */
+  private void compareMaxCleanupTimestampWithTable(HousekeepingMetadata housekeepingMetadata) {
```
massdosage, line 102: To me "compare" suggests a method that just compares but doesn't have any side effects, while this and the one below both change something based on the result of the comparison. Can we rename this method to be clearer, maybe something like `updateCleanupTimestamp` or `setCleanupTimestampToMax`?

beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java
```diff
@@ -79,6 +83,61 @@ private HousekeepingMetadata createOrUpdateHousekeepingMetadata(HousekeepingMeta
...
+
+  /**
+   * When a new partition is added to a table, check to see if the new cleanup delay will be later than the current
+   * cleanup delay for the table.
+   * The cleanup timestamp of a partitioned table should be equivalent to that of the last partition which will be
+   * dropped to prevent premature attempts to cleanup the table.
+   *
+   * @param partitionMetadata
+   */
+  private void comparePartitionCleanupTimestampWithTable(HousekeepingMetadata partitionMetadata) {
```
massdosage, line 122: Similar here, maybe something like `setTableCleanupTimestampToPartitionMax`? (I'm open to other names, my morning coffee hasn't kicked in yet so these may be sub-optimal suggestions ;) )

beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java
```diff
@@ -79,6 +83,61 @@ private HousekeepingMetadata createOrUpdateHousekeepingMetadata(HousekeepingMeta
...
+      tableMetadata.setCleanupTimestamp(partitionCleanupTimestamp);
+      housekeepingMetadataRepository.save(tableMetadata);
+    }
+  }
+
+  private boolean isPartitionedTable(HousekeepingMetadata housekeepingMetadata) {
+    Long numPartitions = housekeepingMetadataRepository.countRecordsForGivenDatabaseAndTableWherePartitionIsNotNull(
+        housekeepingMetadata.getDatabaseName(), housekeepingMetadata.getTableName());
+
+    return numPartitions > 0 && housekeepingMetadata.getPartitionName() == null;
```
massdosage, line 141: Just curious, why do we also need to check that the partition name is null?

massdosage (changes requested):

beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java
JayGreeeen: Yeah you make a good point, will rename them 👍

beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java
JayGreeeen: This method is called when we're updating an existing record and we only want to update a record for a table. So numPartitions tells us if there are any existing partitions on that table, and the null check makes sure the housekeepingMetadata record being updated is for the table and not one of its partitions.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveClient.java
```diff
@@ -100,4 +100,8 @@ public boolean tableExists(String databaseName, String tableName) {
           "Unexpected exception when checking if table \"" + databaseName + "." + tableName + "\" exists.", e);
     }
   }
+
+  public void close(){
```
patduin, line 104: add implement Closeable to the class and you can do try with resources to auto close

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/repository/HousekeepingMetadataRepository.java
```diff
@@ -38,12 +38,20 @@
   @Query(value = "from HousekeepingMetadata t "
       + "where t.databaseName = :databaseName "
       + "and t.tableName = :tableName "
-      + "and (t.partitionName = :partitionName or t.partitionName is NULL) " // To handle special null case
+      + "and (t.partitionName = :partitionName or (:partitionName is NULL and t.partitionName is NULL)) " // To handle special null case
```
patduin, line 41: :partitionName = "xyz"
t.partitionName = "xyz" or ("xyz" is null and t.partitionName is null)

:partitionName = null
t.partitionName = null or (null is null and t.partitionName is null)

I don't get it what problem are you trying to fix here?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/metadata/MetadataCleaner.java
```diff
@@ -19,6 +19,10 @@
 
 public interface MetadataCleaner {
 
+  void init();
```
patduin, line 22: remove both

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/metadata/MetadataCleaner.java
```diff
@@ -19,6 +19,10 @@
...
+
+  void close();
+
   void dropTable(HousekeepingMetadata housekeepingMetadata);
```
patduin, line 26: add client as an argument to both methods

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/ExpiredMetadataHandler.java
```diff
@@ -63,20 +63,10 @@ public ExpiredMetadataHandler(HousekeepingMetadataRepository housekeepingMetadat
...
-    if (dryRunEnabled) {
-      cleanup(housekeepingMetadata, instant, dryRunEnabled);
-    } else {
-      cleanupAndUpdate(housekeepingMetadata, instant, dryRunEnabled);
-    }
-  }
-
-  private void cleanupAndUpdate(HousekeepingMetadata housekeepingMetadata, LocalDateTime instant,
-      boolean dryRunEnabled) {
+    metadataCleaner.init();
```
patduin, line 66: inject the HiveClientFactory hiveClientFactory in the constructor then here you can do:

try (HiveClient client =  hiveClientFactory.newInstance()) {
//body of this whole method

}

patduin (changes requested):
some comment :)

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveMetadataCleaner.java
```diff
@@ -53,6 +62,11 @@ public boolean dropPartition(HousekeepingMetadata housekeepingMetadata) {
 
   @Override
   public boolean tableExists(String databaseName, String tableName) {
-    return client.tableExists(databaseName, tableName);
+    return hiveClient.tableExists(databaseName, tableName);
+  }
+
+  @Override
+  public void close(){
```
nvitucci, line 69: Maybe this one could implement `Closeable` too?

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/repository/HousekeepingMetadataRepository.java
JayGreeeen: when we have `:partitionName = null` anything with that db and table name would get returned because `null is null` is always true; `t.partitionName = "xyz" or null is null`

which breaks the code because we only want 1 value back.

 So it needs to have that check to make sure that when you pass it null you only get the table entry back.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveMetadataCleaner.java
patduin: Nah because I proposed to remove init and close in these classes

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/repository/HousekeepingMetadataRepository.java
patduin: Would splitting this up in two queries make it simpler in the code? on for getting the table record and one for getting the partition records?

JayGreeeen: That is a discussion Vedant and I had when we found the problem, it was either this or have two queries. If you agree this is confusing then I think it is best to separate them out 👍

patduin: it's ok to leave this seems to be the JPA way of doing it :)

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/context/CommonBeans.java
```diff
@@ -81,11 +82,11 @@ public CloseableMetaStoreClientFactory metaStoreClientFactory() {
     return new HiveMetaStoreClientSupplier(metaStoreClientFactory, hiveConf, name);
   }
 
-  @Bean
-  public HiveClient hiveClient(
+  @Bean(name = "hiveClientFactory")
```
patduin, line 85: why name it?

JayGreeeen: Keeping it consistent with the other beans that have seemingly generic methods, but return a specific hive implementation

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/metadata/MetadataCleaner.java
```diff
@@ -19,9 +19,9 @@
 
 public interface MetadataCleaner {
 
-  void dropTable(HousekeepingMetadata housekeepingMetadata);
+  void dropTable(CleanerClient client, HousekeepingMetadata housekeepingMetadata);
```
massdosage, line 22: Summarising some discussions around this change - it's not great to have this coupling here and for the `CleanerClient` to be passed around to all the methods. Having this be an implementation detail of `MetadataCleaner` would be better but there are issues around where the client is created and the amount of logging noise it produces. I think we could solve this all in a nicer way by having some kind of HiveClientPool which manages connections and then the implementation of this could use that to manage connecting to Hive. I'm OK with approving this to move it forward but I'll set up a separate discussion to decide on whether creating a more generic HiveClientPool would be useful here and also potentially in some of our other projects.

massdosage (approved):

nvitucci (approved):
Looks good, although I agree with @massdosage that a solution that doesn't require the client to be passed around could be preferable.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/metadata/MetadataCleaner.java
jmnunezizu: Thanks for summarising that, @massdosage. I was having the same conversation with Jay earlier this morning as I felt the API was not the best with the client being passed in with every interaction.

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperExpiredMetadataSchedulerApiaryIntegrationTest.java
```diff
@@ -147,10 +147,12 @@ public void expiredMetadataAddPartitionEvent() throws SQLException, IOException,
...
     amazonSQS.sendMessage(sendMessageRequest(addPartitionSqsMessage.getFormattedString()));
 
-    await().atMost(TIMEOUT, TimeUnit.SECONDS).until(() -> getExpiredMetadataRowCount() == 1);
+    insertExpiredMetadata("s3://location", null);
+
+    await().atMost(30, TimeUnit.SECONDS).until(() -> getExpiredMetadataRowCount() == 2);
 
     List<HousekeepingMetadata> expiredMetadata = getExpiredMetadata();
-    assertExpiredMetadata(expiredMetadata.get(0), LOCATION_A, PARTITION_A_NAME);
+    assertExpiredMetadata(expiredMetadata.get(1), LOCATION_A, PARTITION_A_NAME);
```
patduin, line 155: what happened to 0?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperExpiredMetadataSchedulerApiaryIntegrationTest.java
```diff
@@ -162,11 +164,13 @@ public void expiredMetadataMultipleAddPartitionEvents() throws SQLException, IOE
...
 
-    await().atMost(TIMEOUT, TimeUnit.SECONDS).until(() -> getExpiredMetadataRowCount() == 2);
+    insertExpiredMetadata("s3://location", null);
+
+    await().atMost(60, TimeUnit.SECONDS).until(() -> getExpiredMetadataRowCount() == 3);
 
     List<HousekeepingMetadata> expiredMetadata = getExpiredMetadata();
-    assertExpiredMetadata(expiredMetadata.get(0), LOCATION_A, PARTITION_A_NAME);
-    assertExpiredMetadata(expiredMetadata.get(1), LOCATION_B, PARTITION_B_NAME);
+    assertExpiredMetadata(expiredMetadata.get(1), LOCATION_A, PARTITION_A_NAME);
```
patduin, line 172: same here we get 3 items now?

patduin: maybe add a assertThat(list.getSize(), is(3));??

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperExpiredMetadataSchedulerApiaryIntegrationTest.java
JayGreeeen: I've had to insert some expired metadata - see line 150, which means theres now 2 entries in the table. This test adds a new partition, but because of the max cleanup timestamp stuff I added there needs to be an existing entry for the table.

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperExpiredMetadataSchedulerApiaryIntegrationTest.java
JayGreeeen: same as above - have inserted an entry for the table on line 167

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/metadata/MetadataCleaner.java
patduin: That sounds good. The old client pool again :).

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperExpiredMetadataSchedulerApiaryIntegrationTest.java
JayGreeeen: An assert would be doing the same as line 169, since it waits till theres 3 rows

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/metadata/MetadataCleaner.java
patduin: we can also just add a factory here in the implementation of this class and create a new client for inside every method. A pool or what we have now is after all just an optimisation detail we might not need it. I think we actually established long time ago that creating a client is fast.

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperExpiredMetadataSchedulerApiaryIntegrationTest.java
patduin: would it make sense to still just assert the get(0) so it's clear what's in the list. It now looks like the test is wrong

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/metadata/MetadataCleaner.java
JayGreeeen: Creating a new client inside every method isnt ideal. I tried it but theres two calls back to back `tableExists` and `dropTable` / `dropPartition` which create two new clients and this adds a lot to the logs which is annoying. As Adrian suggested we could just ignore those logs, but its good to know whats going on with the clients

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperExpiredMetadataSchedulerApiaryIntegrationTest.java
JayGreeeen: Yeah I can add that to make it clearer

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/metadata/MetadataCleaner.java
patduin: ah yes logging can be silenced. All good we can have. cache with a timeout. I think we're all single threaded here so that's fine.

patduin (approved):

PR #65, opened 2020-09-10
Fix tests

JayGreeeen:
I've had to swap the client and housekeepingMetadata around in the `MetadataCleaner` so that we can use the `@TimedTaggable` annotation and assert the tag in the tests. Ideally we will remove the client from the method later when we change how it is created.

### DIFF
--- a/beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporterTest.java
+++ b/beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporterTest.java
@@ -30,6 +30,7 @@ import org.mockito.junit.jupiter.MockitoExtension;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.test.context.support.AnnotationConfigContextLoader;
 
@@ -48,6 +49,7 @@ import com.expediagroup.beekeeper.core.monitoring.Taggable;
 @ContextConfiguration(classes = { TestApplication.class },
   loader = AnnotationConfigContextLoader.class)
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+@TestPropertySource(properties = {"spring.jpa.hibernate.ddl-auto=update"})
 public class BytesDeletedReporterTest {
 
   private static final String TABLE = "database.table";
--- a/beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporterTest.java
+++ b/beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporterTest.java
@@ -30,7 +30,6 @@ import org.mockito.junit.jupiter.MockitoExtension;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ContextConfiguration;
-import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.test.context.support.AnnotationConfigContextLoader;
 
@@ -49,7 +48,6 @@ import com.expediagroup.beekeeper.core.monitoring.Taggable;
 @ContextConfiguration(classes = { TestApplication.class },
   loader = AnnotationConfigContextLoader.class)
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
-@TestPropertySource(properties = {"spring.jpa.hibernate.ddl-auto=update"})
 public class BytesDeletedReporterTest {
 
   private static final String TABLE = "database.table";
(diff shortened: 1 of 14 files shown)


beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporterTest.java
```diff
@@ -48,6 +49,7 @@
 @ContextConfiguration(classes = { TestApplication.class },
   loader = AnnotationConfigContextLoader.class)
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+@TestPropertySource(properties = {"spring.jpa.hibernate.ddl-auto=update"})
```
nvitucci, line 52: Why is this needed here?

JayGreeeen: if its not in here we get a very chatty WARN log; `GenerationTarget encountered exception accepting command : Error executing DDL "drop table housekeeping_metadata" via JDBC Statement`.

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperDryRunMetadataCleanupIntegrationTest.java
```diff
@@ -252,16 +253,21 @@ public void dryRunDontDropPartitionedTable() throws Exception {
   }
 
   @Test
-  public void dryRunMetrics() throws SQLException, TException {
-    hiveTestUtils.createTable(UNPARTITIONED_TABLE_PATH, TABLE_NAME_VALUE, false);
-    amazonS3.putObject(BUCKET, UNPARTITIONED_TABLE_OBJECT_KEY, "");
+  public void dryRunMetrics() throws Exception {
+    Table table = hiveTestUtils.createTable(PARTITIONED_TABLE_PATH, TABLE_NAME_VALUE, true);
```
nvitucci, line 257: Why has the table become partitioned in this test?

JayGreeeen: I changed it so I can check all of the tags are used. For an unpartitioned table it wouldnt use the tag `hive-partition-deleted`, so instead of adding another test for that I changed the table type.

nvitucci: I see. Would it be possible to just add it instead (e.g. `dryRunMetricsPartitioned`), so the unpartitioned table still gets tested? Or would it be a waste?

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporterTest.java
nvitucci: I think it's fine to do it since this is just a test, but out of curiosity: what's the full error?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperDryRunMetadataCleanupIntegrationTest.java
JayGreeeen: can do but it would be testing the same tags - 1, so theres not much point imo. The table still gets tested, this is just specifically for the metrics.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporterTest.java
JayGreeeen: It's a hibernate issue. Seems that the db that gets created is derby and not sql, and it tries to drop the table before creating it, but it doesnt exist so throws the warning. 
`WARN  org.hibernate.tool.schema.internal.ExceptionHandlerLoggedImpl:27 - GenerationTarget encountered exception accepting command : Error executing DDL "drop table housekeeping_metadata" via JDBC Statement
org.hibernate.tool.schema.spi.CommandAcceptanceException: Error executing DDL "drop table housekeeping_metadata" via JDBC Statement
        at org.hibernate.tool.schema.internal.exec.GenerationTargetToDatabase.accept(GenerationTargetToDatabase.java:67)`

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperDryRunMetadataCleanupIntegrationTest.java
nvitucci: Ok then, we can leave it this way.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/monitoring/BytesDeletedReporterTest.java
nvitucci: Interesting, because for Derby the default for `spring.jpa.hibernate.ddl-auto` should be `create-drop`, which should work. Maybe it's overwritten somewhere else.

nvitucci: That's odd. Also, why is it referencing the `housekeeping_metadata` table?

nvitucci (approved):
Looks good.

massdosage (approved):

PR #69, opened 2021-02-02
Changed log error to info in scheduler apiary

javsanbel2:
Really difficult change here, be careful reviewing the PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [3.0.2] - TBD
 ### Fixed
-- Set lifecycle type to `UNREFERENCED` for paths picked up by Vacuum tool. 
+- Set lifecycle type to `UNREFERENCED` for paths picked up by Vacuum tool.
+- Scheduler apiary prints log info instead of log error if the cleanup duration can not be parsed. 
 
 ## [3.0.1] - 2020-09-10
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [3.0.2] - TBD
 ### Fixed
 - Set lifecycle type to `UNREFERENCED` for paths picked up by Vacuum tool.
-- Scheduler apiary prints log info instead of log error if the cleanup duration can not be parsed. 
 
 ## [3.0.1] - 2020-09-10
 ### Fixed
--- a/beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
+++ b/beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
@@ -52,10 +52,9 @@ public class CleanupDelayExtractor {
               listenerEvent.getTableName());
       return value;
     } catch (DateTimeParseException | NullPointerException e) {
-      log
-          .warn(
-              "Overridden delay value '{}' for key '{}' cannot be parsed to a Duration for table '{}.{}'. Using default setting.",
-              tableCleanupDelay, propertyKey, listenerEvent.getDbName(), listenerEvent.getTableName());
+      log.warn(
+              "Overridden delay value '{}' for key '{}' cannot be parsed to a Duration for table '{}.{}'. Using default setting {}.",
+              tableCleanupDelay, propertyKey, listenerEvent.getDbName(), listenerEvent.getTableName(), defaultValue);
       return defaultValue;
     }
   }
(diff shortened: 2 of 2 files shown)


abhimanyugupta07 (approved):

JayGreeeen (approved):

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [3.0.2] - TBD
 ### Fixed
-- Set lifecycle type to `UNREFERENCED` for paths picked up by Vacuum tool. 
+- Set lifecycle type to `UNREFERENCED` for paths picked up by Vacuum tool.
+- Scheduler apiary prints log info instead of log error if the cleanup duration can not be parsed.
```
abhimanyugupta07, line 10: CHANGELOG Not required I think

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
```diff
@@ -48,7 +48,7 @@ public Duration extractCleanupDelay(ListenerEvent listenerEvent) {
     try {
```
patduin, line 48: Sorry but can we change the logic of this class.
And have something like:
```
if tableCleanupDelay is null
  return defaultValue;
else {
try {
      return Duration.parse(tableCleanupDelay);	      return Duration.parse(tableCleanupDelay);
    } catch (DateTimeParseException | NullPointerException e) {	    } catch (DateTimeParseException | NullPointerException e) {
      log.warn(	      log.info(
          "Overridden delay value '{}' for key '{}' cannot be parsed to a Duration for table '{}.{}'. Using default setting.",	          "Overridden delay value '{}' for key '{}' cannot be parsed to a Duration for table '{}.{}'. Using default setting.",
          tableCleanupDelay, propertyKey, listenerEvent.getDbName(), listenerEvent.getTableName());	          tableCleanupDelay, propertyKey, listenerEvent.getDbName(), listenerEvent.getTableName());
      return defaultValue;
}
```
print warn if we found a value that couldn't be parsed. Would also be good to log the value of default that we are actually setting.

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
```diff
@@ -46,11 +46,16 @@ public CleanupDelayExtractor(String propertyKey, String defaultValue) {
...
+      log
+          .info("Using value '{}' for key {} for table {}.{}.", value, propertyKey, listenerEvent.getDbName(),
+              listenerEvent.getTableName());
+      return value;
     } catch (DateTimeParseException | NullPointerException e) {
-      log.error(
-          "Overridden delay value '{}' for key '{}' cannot be parsed to a Duration for table '{}.{}'. Using default setting.",
-          tableCleanupDelay, propertyKey, listenerEvent.getDbName(), listenerEvent.getTableName());
+      log
+          .warn(
```
JayGreeeen, line 56: v picky, but is it the formatter thats splitting `log.info` and `log.warn` onto separate lines? Think it would be nicer on one line

javsanbel2: Do you mean to put the log at the end? So we only use 1 log print statement?

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
```diff
@@ -46,11 +46,16 @@ public CleanupDelayExtractor(String propertyKey, String defaultValue) {
...
+              "Overridden delay value '{}' for key '{}' cannot be parsed to a Duration for table '{}.{}'. Using default setting.",
```
patduin, line 57: `Using default setting '{}'`

Can we print that?

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
JayGreeeen: no I mean have `log.warn` instead of 
```
log
   .warn
```

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
javsanbel2: Yep

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/utils/CleanupDelayExtractor.java
javsanbel2: I can do it, but my format settings (from HCOM) recommend to put it in 2 lines.

patduin (approved):

PR #83, opened 2021-07-08
Github Actions build & release

javsanbel2:
Fixes #


shermosa (approved):

JayGreeeen (approved):

pom.xml
```diff
@@ -33,6 +33,19 @@
     <tag>HEAD</tag>
   </scm>
 
+  <distributionManagement>
```
patduin, line 36: this is already in the parent right: https://github.com/ExpediaGroup/eg-oss-parent/blob/main/pom.xml#L89 Why do you need this.

PR #87, opened 2021-07-15
Add GET /metadata endpoint to API

shermosa:
This PR has the `GET /metadata` endpoint for the Beekeeper-API.

There are some reasons why some code has changed in other modules, these are some examples why:
- I added the Lombok dependency because the `HousekeepingMetadata` and `HousekeepingPath` classes were massive and they are much better with Lombok, this created other issues, like for example some verifications stopped working in the `ExpiredHousekeepingMetadataSchedulerServiceTest.java`.
- To test the [API responses ](https://github.com/ExpediaGroup/beekeeper/blob/0ec1ee46b507963f45c5eaa16b92740c983c20b9/beekeeper-api/src/test/java/com/expediagroup/beekeeper/api/controller/BeekeeperControllerTest.java#L77), there were some JSON objects which were deprecated and I needed to update these dependencies
`    <graphite.version>1.6.5</graphite.version>
    <prometheus.version>1.6.5</prometheus.version>`
which caused more errors in other classes.
- Since the `beekeeper-api` module uses classes from other modules and uses the repositories (example: [HousekeepingMetadataRepository](https://github.com/ExpediaGroup/beekeeper/blob/main/beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/repository/HousekeepingMetadataRepository.java)), I have to add things like `HousekeepingMetadataRepository extends PagingAndSortingRepository<HousekeepingMetadata, Long>`.

I would've only added a new module leaving the others untouched, but sadly since the are connected this wasn't possible. The above doesn't include everything, so if there's any change you'd like to ask about just leave a comment.


CHANGELOG.md
```diff
@@ -3,7 +3,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
-## [3.2.0] - 2021-07-14
+
+## [3.3.0] - TBD
+### Added
+- Added a `GET /metadata` endpoint to the `beekeeper-api`.
+
+## [3.2.0] - TBD
```
vedantchokshi, line 11: Remember to release this before you merge this

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/BeekeeperApiApplication.java
```diff
@@ -0,0 +1,40 @@
...
+@EntityScan(basePackages = { "com.expediagroup.beekeeper.core.model" })
+@EnableJpaRepositories(basePackages = { "com.expediagroup.beekeeper.core.repository" })
+@ComponentScan(basePackages = {
+    "com.expediagroup.beekeeper.api.conf",
+    "com.expediagroup.beekeeper.api.controller",
+    "com.expediagroup.beekeeper.api.service" })
+public class BeekeeperApiApplication {
+  public static void main(String[] args) {
+    new SpringApplicationBuilder(BeekeeperApiApplication.class)
+        .properties("server.port:${endpoint.port:7008}")
```
vedantchokshi, line 36: What is this port number? Remove this line. We will set the port as a property

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/response/HousekeepingMetadataResponse.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.hibernate.annotations.UpdateTimestamp;
+
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+
+import com.expediagroup.beekeeper.core.model.HousekeepingStatus;
+
+/*
```
vedantchokshi, line 32: Maybe stick this comment in the converter and not here? This class will stay even if we have projections. The converter stuff will go

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/response/MetadataResponseConverter.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageImpl;
+
+import com.expediagroup.beekeeper.core.model.HousekeepingMetadata;
+
+public final class MetadataResponseConverter {
```
vedantchokshi, line 26: Can you give this class a private constructor so no objects of the class are made (because they won't do anything as you have static methods)

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/response/MetadataResponseConverter.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private static HousekeepingMetadataResponse convertToHousekeepingMetadataResponse(
```
vedantchokshi, line 28: Put this method below he public method as its used by that method. Gives better readability

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/service/HousekeepingEntityService.java
```diff
@@ -13,11 +13,17 @@
...
 
-import org.springframework.web.bind.annotation.RestController;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.jpa.domain.Specification;
+
+import com.expediagroup.beekeeper.api.response.HousekeepingMetadataResponse;
+import com.expediagroup.beekeeper.core.model.HousekeepingMetadata;
+
+public interface HousekeepingEntityService<HousekeepingEntity> {
```
vedantchokshi, line 25: Can you get rid of the type? You don't need to specify it. Just a plain interface will be fine

beekeeper-api/src/test/java/com/expediagroup/beekeeper/api/response/HousekeepingMetadataResponseTest.java
```diff
@@ -0,0 +1,63 @@
...
+  public void testConvertToHouseKeepingMetadataResponsePage() {
+    HousekeepingMetadata metadata1 = generateDummyHousekeepingMetadata("some_database1", "some_table1");
+    HousekeepingMetadata metadata2 = generateDummyHousekeepingMetadata("some_database2", "some_table2");
+
+    List<HousekeepingMetadata> housekeepingMetadataList = List.of(metadata1, metadata2);
+    Page<HousekeepingMetadataResponse> metadataResponsePage = convertToHousekeepingMetadataResponsePage(
+        new PageImpl<>(housekeepingMetadataList));
+
+    List<HousekeepingMetadataResponse> metadataResponsePageList = metadataResponsePage.getContent();
+    assertThat(metadataResponsePageList.get(0).getDatabaseName()).isEqualTo(metadata1.getDatabaseName());
```
vedantchokshi, line 59: Can we assert the whole object and not just the name?

beekeeper-api/src/test/java/com/expediagroup/beekeeper/api/response/HousekeepingMetadataResponseTest.java
```diff
@@ -0,0 +1,63 @@
...
+  private static final int cleanupAttempts = 0;
+  private static final String lifecycleEventType = EXPIRED.toString();
+
+  @Test
+  public void testConvertToHouseKeepingMetadataResponsePage() {
+    HousekeepingMetadata metadata1 = generateDummyHousekeepingMetadata("some_database1", "some_table1");
+    HousekeepingMetadata metadata2 = generateDummyHousekeepingMetadata("some_database2", "some_table2");
+
+    List<HousekeepingMetadata> housekeepingMetadataList = List.of(metadata1, metadata2);
+    Page<HousekeepingMetadataResponse> metadataResponsePage = convertToHousekeepingMetadataResponsePage(
```
vedantchokshi, line 55: Can we check that the return page has the same configuration has the original page?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
```diff
@@ -0,0 +1,178 @@
...
+import com.expediagroup.beekeeper.core.model.HousekeepingStatus;
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+import com.expediagroup.beekeeper.integration.BeekeeperIntegrationTestBase;
+import com.expediagroup.beekeeper.integration.utils.BeekeeperApiTestClient;
+import com.expediagroup.beekeeper.integration.utils.RestResponsePage;
+
+public class BeekeeperApiIntegrationTest extends BeekeeperIntegrationTestBase {
+
+  public ObjectMapper geObjMapper() {
+    return new ObjectMapper()
```
vedantchokshi, line 62: What are these modules doing?

vedantchokshi: Especially jdk8 given this is all jdk11

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
```diff
@@ -0,0 +1,178 @@
...
+        "s3://some/path/event_date=2020-01-01/event_hour=0/event_type=B",
+        "event_date=2020-01-01/event_hour=0/event_type=B", LifecycleEventType.EXPIRED,
+        Duration.parse("P3D").toString());
+    insertExpiredMetadata(testMetadata1);
+    insertExpiredMetadata(testMetadata2);
+
+    HttpResponse<String> response = testClient.getMetadata();
+    assertThat(response.statusCode()).isEqualTo(OK.value());
+    String body = response.body();
+    assertThrows(ValueInstantiationException.class,
```
vedantchokshi, line 113: What is this? If you are getting an OK status then there should be a valid response from the API?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/utils/BeekeeperApiTestClient.java
```diff
@@ -0,0 +1,54 @@
...
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.net.http.HttpResponse.BodyHandlers;
+
+public class BeekeeperApiTestClient {
+
+  private static final String API_ROOT = "/api/v1";
+  private static final String DATABASE_AND_TABLE = "/database/some_database/table/some_table";
```
vedantchokshi, line 30: Why is this path hardcoded? The db name and table name should be variables that should be put in to this string

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/utils/BeekeeperApiTestClient.java
```diff
@@ -0,0 +1,54 @@
...
+  private static final String METADATA_PATH = "/metadata";
+
+  private final String getHousekeepingEntityUrl;
```
vedantchokshi, line 33: Better name than using a `get` please

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/BeekeeperSchedulerApiary.java
```diff
@@ -42,8 +42,7 @@
   public static void main(String[] args) {
     TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
     new SpringApplicationBuilder(BeekeeperSchedulerApiary.class)
-        .properties("spring.config.additional-location:classpath:/beekeeper-scheduler-apiary-application.yml,"
-            + "${config:null}")
+        .properties("spring.config.additional-location:classpath:/beekeeper-scheduler-apiary-application.yml")
```
vedantchokshi, line 45: You removed too much. It needs the `${config:/}` instead

beekeeper-scheduler/src/test/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerServiceTest.java
```diff
@@ -123,13 +120,9 @@ public void typicalUpdatePartitionedTableWithShorterCleanupDelay() {
     verify(existingTable).setHousekeepingStatus(metadata.getHousekeepingStatus());
     verify(existingTable).setCleanupDelay(metadata.getCleanupDelay());
     verify(existingTable).setClientId(metadata.getClientId());
-    verify(existingTable).getCleanupTimestamp();
```
vedantchokshi, line 126: Why have you removed these verifications in the scheduler?

vedantchokshi (changes requested):
Few comments and questions but getting there 😄

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/BeekeeperApiApplication.java
shermosa: Because if not, it will happen in 8080 which is the default, and in terraform I need to specify the port anyway, so I thought might as well do it on another one that's not the default

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
shermosa: They set the environment, I removed the jdk8 one because it wasn't necessary, the other two are

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
shermosa: Because when a table is not found, what is displayed is an empty list. Therefore there's not an error, the API still works, so the status is still OK

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/utils/BeekeeperApiTestClient.java
shermosa: Since this is used for testing, I always use the same database name and table name, as the only time I would need a different one is to check what happens when a table is not found, and this is done here: https://github.com/ExpediaGroup/beekeeper/blob/a225d25ecd2f7f56f9af98ce4eb286d313ce2d51/beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java#L99

beekeeper-scheduler/src/test/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerServiceTest.java
shermosa: Because when implementing Lombok, some of these verifications stopped working. I asked you what to do about them and you said that it was fine to delete them

beekeeper-api/src/test/java/com/expediagroup/beekeeper/api/response/HousekeepingMetadataResponseTest.java
shermosa: Yes good call, I added an assertion https://github.com/ExpediaGroup/beekeeper/blob/21ce01c66ba2422af956885f9a23be74ebb6562d/beekeeper-api/src/test/java/com/expediagroup/beekeeper/api/response/HousekeepingMetadataResponseTest.java#L45

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/response/MetadataResponseConverter.java
```diff
@@ -0,0 +1,67 @@
...
+/*
+Ideally, this should be done with Spring projections (See https://www.baeldung.com/spring-data-jpa-projections)
+but currently projections do not support using Specification (See https://github.com/spring-projects/spring-data-jpa/issues/1378).
+
+Therefore, this is a 'manual' conversion of the HousekeepingMetadata object, which will be used in the API.
+*/
+
+public final class MetadataResponseConverter {
+
+  private MetadataResponseConverter(){};
```
vedantchokshi, line 35: Needs space after brackets. Can you make you format all the files that you're changing?

beekeeper-api/src/test/java/com/expediagroup/beekeeper/api/response/HousekeepingMetadataResponseTest.java
```diff
@@ -0,0 +1,53 @@
...
+        new PageImpl<>(housekeepingMetadataList));
+
+    List<HousekeepingMetadataResponse> metadataResponsePageList = metadataResponsePage.getContent();
+
+    assertThat(metadataResponsePageList.get(0)).isEqualTo(metadataResponse1);
+    assertThat(metadataResponsePageList.get(1)).isEqualTo(metadataResponse2);
+    assertTrue(metadataResponsePage.getPageable().equals((new PageImpl<>(housekeepingMetadataList).getPageable())));
```
vedantchokshi, line 50: Change this to assertThat too

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperIntegrationTestBase.java
```diff
@@ -262,4 +278,17 @@ private HousekeepingMetadata createHousekeepingMetadata(
         .build();
   }
 
+  protected void assertThatMetadataEqualsResponse(
```
vedantchokshi, line 281: Don't put this in the base. This is API specific

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperIntegrationTestBase.java
```diff
@@ -240,13 +255,14 @@ private HousekeepingPath createHousekeepingPath(String path, LifecycleEventType
         .build();
   }
 
-  private HousekeepingMetadata createHousekeepingMetadata(
+  protected HousekeepingMetadata createHousekeepingMetadata(
```
vedantchokshi, line 258: Why did this change to protected?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
vedantchokshi: Okay. If you're using it to make a bean in the source code then can you make sure you change it there too.

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
vedantchokshi: As you should just read the value and the content length should be zero right? Not sure where the exception comes in

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/utils/BeekeeperApiTestClient.java
vedantchokshi: But for future proofing and just because this is a client, can you make it generic please?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
shermosa: Because it throws this exception: `Cannot construct instance of 
 com.expediagroup.beekeeper.integration.utils.RestResponsePage, problem: Page size must not be less than one!`, so if this exception is thrown it means the page size is 0 and no tables were found, which is what we want

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/BeekeeperIntegrationTestBase.java
shermosa: Because I use this method in `BeekeeperApiIntegrationTest` class too, where my integration tests for the API are

vedantchokshi: This is an internal method. There are are other methods that allow you to insert metadata. Why can't you use them?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
vedantchokshi: But that's a generic error jackson throws when it can't deserialize the object. You need to figure out a way to parse a the page as an empty. The main that's causing the error is `PageRequest.of(number, size)`. Size needs to be greater than 0. Maybe make a new object for empty page or modify this one.

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/utils/BeekeeperApiTestClient.java
```diff
@@ -0,0 +1,54 @@
...
+  private final HttpClient httpClient;
+
+  public BeekeeperApiTestClient(String baseUrl) {
+    this.housekeepingEntityUrl = baseUrl + API_ROOT;
+    this.httpClient = HttpClient.newHttpClient();
+  }
+
+  public HttpResponse<String> getMetadata(String database, String table) throws IOException, InterruptedException {
+    HttpRequest request = newBuilder()
+        .uri(URI.create(housekeepingEntityUrl + "/database/" + database + "/table/" + table + METADATA_PATH))
```
vedantchokshi, line 42: Can we use string.format() here instead?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/utils/BeekeeperApiTestClient.java
```diff
@@ -0,0 +1,54 @@
...
+        .GET()
+        .build();
+    return httpClient.send(request, BodyHandlers.ofString());
+  }
+
+  public HttpResponse<String> getMetadata(String database, String table, String filters)
+    throws IOException, InterruptedException {
+    HttpRequest request = newBuilder().uri(URI.create(housekeepingEntityUrl + "/database/" + database + "/table/" + table + METADATA_PATH + filters)).GET().build();
```
vedantchokshi, line 50: Here too

vedantchokshi (approved):
😄

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [3.3.0] - TBD
```
jmnunezizu, line 7: I believe we agreed to use Unreleased rather than TBD. Would you mind changing it, please? Thanks.

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/BeekeeperApiApplication.java
jmnunezizu: Then you could leave it on 8080. If you chose a different port you'd probably have to document that unless it's specified, the application will start on port 7008 which is quite an odd selection.

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/controller/BeekeeperController.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.expediagroup.beekeeper.api.response.HousekeepingMetadataResponse;
+import com.expediagroup.beekeeper.api.service.HousekeepingEntityServiceImpl;
+import com.expediagroup.beekeeper.core.model.HousekeepingMetadata;
+
+@RequestMapping("/api/v1")
+@RestController
+public class BeekeeperController {
+
+  private final HousekeepingEntityServiceImpl housekeepingEntityServiceImpl;
```
jmnunezizu, line 42: Your reference should point to the interface and not to the concrete class (https://github.com/shermosa/beekeeper/blob/metadata/beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/service/HousekeepingEntityService.java). Always program to interfaces.

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/controller/BeekeeperController.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Autowired
+  public BeekeeperController(HousekeepingEntityServiceImpl housekeepingEntityServiceImpl) {
```
jmnunezizu, line 45: The same thing here, you should use an interface reference, not a concrete class.

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/controller/BeekeeperController.java
jmnunezizu: Also, you could probably change the name of the property to be `housekeepingEntityService` and drop the `Impl` from it.

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/service/HousekeepingEntityServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.jpa.domain.Specification;
+import org.springframework.stereotype.Service;
+
+import com.expediagroup.beekeeper.api.response.HousekeepingMetadataResponse;
+import com.expediagroup.beekeeper.core.model.HousekeepingMetadata;
+import com.expediagroup.beekeeper.core.repository.HousekeepingMetadataRepository;
+
+@Service
+public class HousekeepingEntityServiceImpl implements HousekeepingEntityService{
```
jmnunezizu, line 31: There's a space missing before the opening curly brace.

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+  public ObjectMapper geObjMapper() {
+    return new ObjectMapper()
+        .registerModule(new ParameterNamesModule())
+        .registerModule(new JavaTimeModule());
+  }
+
+  private static final Logger log = LoggerFactory.getLogger(BeekeeperApiIntegrationTest.class);
+
+  // APP CONTEXT AND TEST CLIENT
```
jmnunezizu, line 71: I think this comment is not adding any value.

pom.xml
```diff
@@ -42,12 +42,12 @@
     <docker.registry>expediagroup</docker.registry>
     <jdk.version>11</jdk.version>
     <junit.jupiter.version>5.6.0</junit.jupiter.version>
-    <logback.version>1.2.3</logback.version>
+    <logback.version>1.2.3</logback.version>    <lombok.version>1.18.20</lombok.version>
```
jmnunezizu, line 45: The lombok dependency looks out of place.

jmnunezizu (changes requested):

beekeeper-api/src/main/java/com/expediagroup/beekeeper/api/controller/BeekeeperController.java
```diff
@@ -0,0 +1,69 @@
...
+import com.expediagroup.beekeeper.core.model.HousekeepingMetadata;
+
+@RequestMapping("/api/v1")
+@RestController
+public class BeekeeperController {
+
+  private final HousekeepingEntityService housekeepingEntityService;
+
+  @Autowired
+  public BeekeeperController(HousekeepingEntityServiceImpl housekeepingEntityService) {
```
jmnunezizu, line 46: The reference is still using the concrete class rather than the interface. It should be changed to
```
public BeekeeperController(HousekeepingEntityService houskeepingEntityService) {
  ...
}
```

beekeeper-api/src/test/java/com/expediagroup/beekeeper/api/TestApplication.java
```diff
@@ -16,16 +16,10 @@
...
-  public static void main(String[] args) {
-    new SpringApplicationBuilder(BeekeeperApi.class)
-        .properties("server.port:${endpoint.port:7008}")
-        .build()
-        .run(args);
-  }
-
-}
+@EnableConfigurationProperties
+@ComponentScan(basePackages = { "com.expediagroup.beekeeper.api.conf", "com.expediagroup.beekeeper.api.controller" })
```
patduin, line 24: This can go right?

beekeeper-integration-tests/src/test/java/com/expediagroup/beekeeper/integration/api/BeekeeperApiIntegrationTest.java
```diff
@@ -0,0 +1,214 @@
...
+import com.expediagroup.beekeeper.core.model.HousekeepingMetadata;
+import com.expediagroup.beekeeper.core.model.HousekeepingStatus;
+import com.expediagroup.beekeeper.core.model.LifecycleEventType;
+import com.expediagroup.beekeeper.integration.BeekeeperIntegrationTestBase;
+import com.expediagroup.beekeeper.integration.utils.BeekeeperApiTestClient;
+import com.expediagroup.beekeeper.integration.utils.RestResponsePage;
+
+public class BeekeeperApiIntegrationTest extends BeekeeperIntegrationTestBase {
+
+  public ObjectMapper geObjMapper() {
```
jmnunezizu, line 63: Minor, but you might want to rename this to `getObjectMapper` instead.

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/ExpiredHousekeepingMetadataGenerator.java
```diff
@@ -119,7 +119,7 @@ private HousekeepingEntity generateHousekeepingEntity(AlterPartitionEvent listen
   private HousekeepingEntity generateHousekeepingEntity(ListenerEvent listenerEvent, String clientId, String path,
       String partitionName) {
     Duration cleanupDelay = cleanupDelayExtractor.extractCleanupDelay(listenerEvent);
-    HousekeepingMetadata.Builder builder = new HousekeepingMetadata.Builder()
+    HousekeepingMetadata.HousekeepingMetadataBuilder builder = HousekeepingMetadata.builder()
```
jmnunezizu, line 122: You don't need a reference to the builder here. You could replace it with
```
return HousekeepingMetadata.builder()
 ...
 .build();
```

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/generator/UnreferencedHousekeepingPathGenerator.java
```diff
@@ -100,7 +100,7 @@ public LifecycleEventType getLifecycleEventType() {
   public HousekeepingEntity generateHousekeepingEntity(ListenerEvent listenerEvent, String clientId,
       String cleanupPath) {
     Duration cleanupDelay = cleanupDelayExtractor.extractCleanupDelay(listenerEvent);
-    HousekeepingPath.Builder builder = new HousekeepingPath.Builder()
+    HousekeepingPath.HousekeepingPathBuilder builder = HousekeepingPath.builder()
```
jmnunezizu, line 103: Same as previous comment.

jmnunezizu (approved):

beekeeper-api/src/test/java/com/expediagroup/beekeeper/api/TestApplication.java
shermosa: yes

jmnunezizu (approved):

patduin (approved):

JayGreeeen (approved):

PR #93, opened 2021-07-21
Add profile from parent adding gpgArguments

javsanbel2:
I added the profile from the pom parent but adding a gpgConfiguration. I saw from different places that this maybe resolve the issue:

Links:

- Molten repository from Expedia Group: https://github.com/ExpediaGroup/molten/blob/master/pom.xml#L166
- First stack overflow issue: https://stackoverflow.com/questions/53992950/maven-gpg-plugin-failing-with-inappropriate-ioctl-for-device-when-running-unde
- Same issue but in Github Actions: https://github.com/samuelmeuli/action-maven-publish/issues/1

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
 
   <profiles>
     <profile>
-      <id>sonatype-oss-release-beekeper</id>
+      <id>sonatype-oss-release-beekepeer</id>
       <build>
         <plugins>
           <plugin>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -160,4 +160,48 @@
       </plugins>
     </pluginManagement>
   </build>
+
+  <profiles>
```
javsanbel2, line 164: Maybe I should not add it here and add this configuration inside the <plugin>

patduin (approved):
let's try if this works we need to maybe make a PR to add this to the parent pom

shermosa (approved):

PR #94, opened 2021-07-21
Remove skip.tests from release.yml

shermosa:
Fixes #

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -160,48 +160,4 @@
       </plugins>
     </pluginManagement>
   </build>
-
-  <profiles>
-    <profile>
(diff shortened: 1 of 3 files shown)


javsanbel2 (approved):

javsanbel2:
Update parent version

pom.xml
```diff
@@ -160,48 +160,4 @@
       </plugins>
     </pluginManagement>
   </build>
-
```
patduin, line 163: you need to update the parent pom still right

patduin (changes requested):
blocked by: https://github.com/ExpediaGroup/eg-oss-parent/pull/19

PR #96, opened 2021-08-12
Updated snakeyaml dependency

shermosa:
Updated the `snakeyaml` dependency to `1.27` due to an error in kubernetes where it was trying to find a method which didn't exist in version `1.24`.


JayGreeeen (approved):

beekeeper-api/pom.xml
```diff
@@ -36,6 +36,10 @@
       <groupId>org.projectlombok</groupId>
       <artifactId>lombok</artifactId>
     </dependency>
+    <dependency>
+      <groupId>mysql</groupId>
+      <artifactId>mysql-connector-java</artifactId>
```
javsanbel2, line 41: Why this dependency is needed?

shermosa: Because I was getting a 'Driver not found' error in k8s in the beekeeper-api module and it was solved by adding this dependency. For more information look here https://stackoverflow.com/questions/52804228/failed-to-load-driver-class-com-mysql-jdbc-driver

javsanbel2 (approved):

PR #107, opened 2021-11-08
Add index for `housekeeping_path` table

andreeapad:


beekeeper-scheduler-apiary/src/main/resources/db/migration/V2_2__Add_index_for_housekeeping_path.sql
```diff
@@ -0,0 +1,3 @@
+USE beekeeper;
+
+ALTER TABLE housekeeping_path ADD INDEX `table_index` (`housekeeping_status`, `database_name`(300), `table_name`(300));
```
abhimanyugupta07, line 3:
```suggestion
ALTER TABLE housekeeping_path ADD INDEX `housekeeping_path_index` (`housekeeping_status`, `database_name`(300), `table_name`(300));
```

abhimanyugupta07:
I am thinking, we can add the index for the other beekeeper table too so we don't run into same issues with that later.

abhimanyugupta07 (approved):

patduin (approved):

PR #121, opened 2021-12-15
Fix disable operation

andreeapad:


patduin (approved):

javsanbel2 (approved):

javsanbel2:
What about the versions? Is that okay? It was wrong but deployed correctly? or do we need to release again in those versions

andreeapad:
the changelog was wrong, i added the versions that were actually released

abhimanyugupta07 (approved):

PR #129, opened 2022-01-12
Disable entries for deleted tables

andreeapad:


abhimanyugupta07 (approved):

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveClient.java
```diff
@@ -117,6 +117,9 @@ public boolean tableExists(String databaseName, String tableName) {
         return new HashMap<>();
       }
       return table.getParameters();
```
javsanbel2, line 119: could you put this at the end please? it took me a bit of time to realise about structure

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveClient.java
```diff
@@ -117,6 +117,9 @@ public boolean tableExists(String databaseName, String tableName) {
...
+    } catch (NoSuchObjectException e) {
+      log.warn("The table {}.{} does not exist", databaseName, tableName);
```
javsanbel2, line 121: Maybe a bit more informative this log? Could we say that we are going to remove table properties?

andreeapad: this is a general hive util method, so it may not apply to other calls

javsanbel2 (approved):

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/hive/HiveClient.java
abhimanyugupta07: what was the resolution about this ? I am curious :)

PR #145, opened 2022-09-28
Feature/fix s3 files not deleted

RickArt:
Fixes #
Added logic to break down s3 file deleting request on smaller ones to avoid an exception for large (over 1000 files) request.


beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
```diff
@@ -207,6 +207,21 @@ void deleteObjectsInDirectory() {
     assertThat(amazonS3.doesObjectExist(bucket, key2)).isFalse();
   }
 
+  @Test
+  void deleteThousandObjectsInDirectory() {
```
javsanbel2, line 211: Could you add in this test an assertion where if you try to delete the same quantity of objects using the s3 service provided by the aws library the amount of deleted objects are 1000? (instead of 1234)

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
```diff
@@ -73,10 +74,21 @@ List<String> deleteObjects(String bucket, List<String> keys) {
       return Collections.emptyList();
     }
     if (!dryRunEnabled) {
-      keys.forEach(key -> log.info("Deleting: \"{}/{}\"", bucket, key));
-      DeleteObjectsRequest deleteObjectsRequest = new DeleteObjectsRequest(bucket)
-          .withKeys(keys.toArray(new String[] {}));
-      DeleteObjectsResult deleteObjectsResult = amazonS3.deleteObjects(deleteObjectsRequest);
+      log.info("Attempting to delete a total of {} objects, from [{}] to [{}]", keys.size(), keys.get(0), keys.get(keys.size()-1));
```
jmnunezizu, line 77: You need to leave a space before and after `-`. You have some other cosmetic issues with this change too. Please check with the team and run this change through our formatter to ensure there are no issues.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
```diff
@@ -73,10 +74,21 @@ List<String> deleteObjects(String bucket, List<String> keys) {
...
+      DeleteObjectsResult deleteObjectsResult = new DeleteObjectsResult(new ArrayList<>());
+      DeleteObjectsRequest deleteObjectsRequest = new DeleteObjectsRequest(bucket);
+      int totalKeys = keys.size();
+      int chunkSize = 1000;
+      int indexStart;
+      int indexEnd = 0;
+      while(indexEnd < totalKeys) {
+        indexStart = indexEnd;
+        indexEnd = indexStart + chunkSize < totalKeys ? indexStart + chunkSize : totalKeys;
```
jmnunezizu, line 86: It'd be better to give `indexStart + chunkSize` a variable name that has a meaning. Right now, I don't know why the sum of those needs to be smaller than `totalKeys`.
Also, you could consider adding this into a private function that houses the logic.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
```diff
@@ -73,10 +74,21 @@ List<String> deleteObjects(String bucket, List<String> keys) {
...
+        deleteObjectsRequest .withKeys(keys.subList(indexStart, indexEnd).toArray(new String[] {}));
```
jmnunezizu, line 87: What happens if one of these requests fails? Is it retried?

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
```diff
@@ -207,6 +207,21 @@ void deleteObjectsInDirectory() {
     assertThat(amazonS3.doesObjectExist(bucket, key2)).isFalse();
   }
 
+  @Test
```
jmnunezizu, line 210: You should also add a test to see what happens when one of the delete operations fails.

jmnunezizu (changes requested):

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
```diff
@@ -207,6 +207,21 @@ void deleteObjectsInDirectory() {
...
+    int totalObjects = 1234;
+    for(int i = 1; i <= totalObjects; i++){
+      var key = key1+i;
+      amazonS3.putObject(bucket, key, content);
+      keys.add(key);
+    }
+
+    List<String> result = s3Client.deleteObjects(bucket, keys);
+
+    assertThat(result.size()).isEqualTo(totalObjects);
```
andreeapad, line 222: you can probably assert the actual number of objects left by doing a get objects s3 call

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
RickArt: At the moment there is no retry logic for this part of beekeeper, I see on some other part of the project we are using spring retry library, that could be an option. Alternatively i think i can add some specific retry logic by further breaking the request on smaller parts trying to delete as many files as possible and maybe find out if a specific file is failing to be deleted.

jmnunezizu: Thanks, @RickArt. I think we should raise another issue to look at retry logic. As you said, there might be other parts that already have some retry mechanism in place.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
RickArt: Changed the test to work with 1000 files instead of 1234

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
RickArt: fixed space and checked style formatter

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
RickArt: moved the logic to another function and added some names to make it more readable

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
RickArt: added the assertion

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
RickArt: created a new ticket to track the feature

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
RickArt: will work on this on the new ticket

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
```diff
@@ -73,10 +73,21 @@ List<String> deleteObjects(String bucket, List<String> keys) {
       return Collections.emptyList();
     }
     if (!dryRunEnabled) {
-      keys.forEach(key -> log.info("Deleting: \"{}/{}\"", bucket, key));
-      DeleteObjectsRequest deleteObjectsRequest = new DeleteObjectsRequest(bucket)
-          .withKeys(keys.toArray(new String[] {}));
-      DeleteObjectsResult deleteObjectsResult = amazonS3.deleteObjects(deleteObjectsRequest);
+      log.info("Attempting to delete a total of {} objects, from [{}] to [{}]", keys.size(), keys.get(0), keys.get(keys.size() - 1));
```
alangr28, line 76: Logs are different. Is this expected?

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
```diff
@@ -73,10 +73,21 @@ List<String> deleteObjects(String bucket, List<String> keys) {
...
+      DeleteObjectsResult deleteObjectsResult = new DeleteObjectsResult(new ArrayList<>());
+      DeleteObjectsRequest deleteObjectsRequest = new DeleteObjectsRequest(bucket);
+      int totalKeys = keys.size();
+      int chunkSize = 1000;
```
alangr28, line 80: It would be great if we can turn this into a constant.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
RickArt: logs were having a negative effect on tracking issues, first, it stated it was deleting files before actually deleting them, which was misleading at the moment of tracking the error, it also happened to add too many logs that rendered previous logs to be lost on the pods, so i thought of making them shorter and more useful

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/ExpiredMetadataHandler.java
```diff
@@ -98,13 +98,14 @@ private boolean cleanup(CleanerClient client, HousekeepingMetadata housekeepingM
...
 
+
   private void cleanUpTable(CleanerClient client, HousekeepingMetadata housekeepingMetadata) {
     String databaseName = housekeepingMetadata.getDatabaseName();
     String tableName = housekeepingMetadata.getTableName();
     log.info("Cleaning up metadata for \"{}.{}\"", databaseName, tableName);
     if (metadataCleaner.tableExists(client, databaseName, tableName)) {
-      metadataCleaner.dropTable(housekeepingMetadata, client);
       pathCleaner.cleanupPath(housekeepingMetadata);
+      metadataCleaner.dropTable(housekeepingMetadata, client);
```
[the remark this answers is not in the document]
RickArt: Because it ended up dropping meta data before actually dropping the files, which was misleading when tracking the problem, because looking at hive meta store, the table wasn't listed anymore. so i thought this would help better reflect what the process is actually doing

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
```diff
@@ -207,6 +207,25 @@ void deleteObjectsInDirectory() {
...
+  void deleteThousandObjectsInDirectory() {
+    ArrayList<String> keys = new ArrayList<>();
+    int totalObjects = 1000;
+    for(int i = 1; i <= totalObjects; i++){
+      var key = keyRoot + "/file" + i;
+      keys.add(key);
+    }
+    keys.parallelStream().forEach(key -> amazonS3.putObject(bucket, key, content));
+    int numberOfObjectsLeft = amazonS3.listObjects(bucket, keyRoot).getObjectSummaries().size();
+    assertThat(numberOfObjectsLeft).isEqualTo(totalObjects);
```
andreeapad, line 220: i think these lines can be removed cause we don't need to test the s3 `putObject` functionality

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
```diff
@@ -207,6 +207,25 @@ void deleteObjectsInDirectory() {
...
+
+    List<String> result = s3Client.deleteObjects(bucket, keys);
+    numberOfObjectsLeft = amazonS3.listObjects(bucket, keyRoot).getObjectSummaries().size();
+
+    assertThat(result.size()).isEqualTo(totalObjects);
```
andreeapad, line 225: can you move this assertion a line above? numberOfObjectsLeft can be moved below

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
```diff
@@ -207,6 +207,25 @@ void deleteObjectsInDirectory() {
...
+    int totalObjects = 1000;
+    for(int i = 1; i <= totalObjects; i++){
+      var key = keyRoot + "/file" + i;
+      keys.add(key);
+    }
+    keys.parallelStream().forEach(key -> amazonS3.putObject(bucket, key, content));
+    int numberOfObjectsLeft = amazonS3.listObjects(bucket, keyRoot).getObjectSummaries().size();
+    assertThat(numberOfObjectsLeft).isEqualTo(totalObjects);
+
+    List<String> result = s3Client.deleteObjects(bucket, keys);
```
andreeapad, line 222: i saw the check for the number of keys deleted, but you should check that the names of the keys are correct too (not sure if the order matters too..)

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/ExpiredMetadataHandler.java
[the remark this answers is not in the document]
andreeapad: @abhimanyugupta07 not sure whether there are implications for the delete order here 🤔

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
andreeapad: if you only have 1000 objects, then the loop will only be executed once. you should add a test where the execution is repeated for at least 2 runs. (at least 1001 object i think)

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
andreeapad: i agree, the logs may be more useful in this new format

beekeeper-metadata-cleanup/src/main/java/com/expediagroup/beekeeper/metadata/cleanup/handler/ExpiredMetadataHandler.java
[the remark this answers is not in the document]
RickArt: restored the deletion order

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
RickArt: true, removed the assertion

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
RickArt: I don't think we need to check that since we are creating the files from the `keys` list, which is the same used for the deletion.

andreeapad: the keys list returned from that method is created with logic that we wrote, so it should be verified too

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
```diff
@@ -73,10 +74,20 @@ List<String> deleteObjects(String bucket, List<String> keys) {
...
+      log.info("Attempting to delete a total of {} objects, from [{}] to [{}]", keys.size(), keys.get(0), keys.get(keys.size() - 1));
+      DeleteObjectsResult deleteObjectsResult = new DeleteObjectsResult(new ArrayList<>());
+      DeleteObjectsRequest deleteObjectsRequest = new DeleteObjectsRequest(bucket);
+      int totalKeys = keys.size();
+      int indexStart;
+      int indexEnd = 0;
+      while(indexEnd < totalKeys) {
+        indexStart = indexEnd;
+        indexEnd = nextIndexEnd(indexStart, REQUEST_CHUNK_SIZE, totalKeys);
+        deleteObjectsRequest.withKeys(keys.subList(indexStart, indexEnd).toArray(String[]::new));
```
andreeapad, line 86: this line builds the returned keys, so this is what i'm suggesting to verify in the test

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
```diff
@@ -73,10 +74,20 @@ List<String> deleteObjects(String bucket, List<String> keys) {
...
+        deleteObjectsResult.getDeletedObjects().addAll(
```
andreeapad, line 87: here we modify the to-be-returned keys, so this is something that could be verified too (referring to the comment about the unit test)

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
RickArt: added assertion

andreeapad (approved):

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
jmnunezizu: Thanks. You could link it here for future reference.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3BytesDeletedCalculatorTest.java
```diff
@@ -38,7 +38,7 @@ class S3BytesDeletedCalculatorTest {
 
   private long contentBytes = "content".getBytes().length;
   private String bucket = "bucket";
-  private String key1 = "db/table/id/partition1/file1";
+  private String key1 = "apiary-112945290135-us-east-1-user/mass/stream/output/segment_id=628b64be-188a-4d23-8e93-0dee521e8321/run_dt=2022-06-18";
```
abhimanyugupta07, line 41: remove the account numbers and real paths from here as this is open source.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3ClientTest.java
RickArt: Merged the tests and changed the it to be parameterized

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
RickArt: The created ticket is [EGDL-3574](https://jira.expedia.biz/browse/EGDL-3574)

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
```diff
@@ -73,10 +74,20 @@ List<String> deleteObjects(String bucket, List<String> keys) {
...
-      DeleteObjectsRequest deleteObjectsRequest = new DeleteObjectsRequest(bucket)
-          .withKeys(keys.toArray(new String[] {}));
-      DeleteObjectsResult deleteObjectsResult = amazonS3.deleteObjects(deleteObjectsRequest);
+      log.info("Attempting to delete a total of {} objects, from [{}] to [{}]", keys.size(), keys.get(0), keys.get(keys.size() - 1));
+      DeleteObjectsResult deleteObjectsResult = new DeleteObjectsResult(new ArrayList<>());
+      DeleteObjectsRequest deleteObjectsRequest = new DeleteObjectsRequest(bucket);
+      int totalKeys = keys.size();
+      int indexStart;
+      int indexEnd = 0;
+      while(indexEnd < totalKeys) {
```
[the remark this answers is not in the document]
RickArt: changed it, and ran code formatter on the set of changes for this file

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3BytesDeletedCalculatorTest.java
RickArt: I forgot about this, it was a test but i missed returning it back.

beekeeper-cleanup/src/main/java/com/expediagroup/beekeeper/cleanup/aws/S3Client.java
jmnunezizu: @RickArt – I deleted your comment. You can't link our tickets here because this is open source. You need to create an issue here in GitHub to track it and get it implemented.

beekeeper-cleanup/src/test/java/com/expediagroup/beekeeper/cleanup/aws/S3BytesDeletedCalculatorTest.java
jmnunezizu: @RickArt this information needs to disappear ASAP.

PR #166, opened 2023-09-14
Fix for trailing /

patduin:


beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/LocationNormalizer.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Attempts to normalize string representing a location of a Hive Table, could have a wide variety of schemes, s3, s3a,
+ * hdfs etc..
+ */
+public class LocationNormalizer {
+
+  public String normalize(String location) {
```
patduin, line 26: I think we can expand on this in follow up PR to normalize scheme s3a/s3n/s3 to s3

beekeeper-scheduler-apiary/src/main/java/com/expediagroup/beekeeper/scheduler/apiary/filter/LocationOnlyUpdateListenerEventFilter.java
```diff
@@ -44,6 +55,11 @@ public boolean isFiltered(ListenerEvent listenerEvent, LifecycleEventType lifecy
   }
 
   private boolean isLocationSame(String oldLocation, String location) {
-    return location == null || oldLocation == null || oldLocation.equals(location);
+    if (location == null || oldLocation == null) {
+      return true;
+    }
+    String normalizedOldLocation = locationNormalizer.normalize(oldLocation);
+    String normalizedLocation = locationNormalizer.normalize(location);
+    return normalizedOldLocation.equals(normalizedLocation);
```
patduin, line 63: We should expand in follow up PR to check that location doesn't start with oldLocation and vica versa.

old: `s3://foo/bar`
new: `s3:/foo`

Should be considered same.

jmnunezizu: I can't say right now if it's a good idea.

beekeeper-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/apiary/filter/LocationNormalizerTest.java
```diff
@@ -0,0 +1,45 @@
...
+    assertThat(normalizer.normalize(null)).isEqualTo(null);
+    assertThat(normalizer.normalize("foo")).isEqualTo("foo");
+    assertThat(normalizer.normalize("foo/bar")).isEqualTo("foo/bar");
+  }
+  
+  @Test
+  void normalizeTrailingSlash() {
+    assertThat(normalizer.normalize("s3://bucket/prefix/")).isEqualTo("s3://bucket/prefix");
+    assertThat(normalizer.normalize("s3://bucket/prefix///")).isEqualTo("s3://bucket/prefix");
+    assertThat(normalizer.normalize("s3://bucket/prefix/../")).isEqualTo("s3://bucket/prefix/..");
```
jmnunezizu, line 41: This looks wrong in a traditional file system. 
`s3://bucket/prefix/../` is equal to `s3://bucket`.

jmnunezizu (commented):
Please check the comment about `/../`.

beekeeper-core/src/main/java/com/expediagroup/beekeeper/core/model/PeriodDuration.java
andreeapad, line 1: the licence header is different because this class was copied from https://github.com/ThreeTen/threeten-extra/blob/main/src/main/java/org/threeten/extra/PeriodDuration.java

patduin: ah ok these shouldn't be changed at all I thought it was just copyright, we should exlude them I'll have a look

beekeeper-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/apiary/filter/LocationNormalizerTest.java
patduin: yeah I'll remove the example

patduin: I agree but the "algorithm" just does / removal. I'm not even sure you can have relative paths in hfds (s3 for sure not). 
There are probably more usecase where you can specify a path in multiple ways pointing to the same thing.
My example was made up I'm not sure if it's relevant (at this point).

jmnunezizu: For future discussion: https://issues.apache.org/jira/browse/HDFS-836

jmnunezizu (approved):

andreeapad (approved):

PR #181, opened 2024-11-11
Initial Commit to fix actions on iceberg tables

HamzaJugon:
Fixes #


HamzaJugon:
Added predicate, events filter & updated scheduler.
Need to verify if tableParameters is enough to go by when verifying if it is an Iceberg table or not. In IceKeeper we use both storageDescriptor and TableParameters.

HamzaJugon:
Updated to allow identification of Iceberg tables by using both descriptor properties and table parameters (`table_type`) 
It now accepts a `Table` instead of a Map.

PR #192, opened 2025-05-07
Use partition creation time for TTL scheduling of Hive partitions

HamzaJugon:
Fixes #

### DIFF
--- a/beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/hive/PartitionInfo.java
+++ b/beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/hive/PartitionInfo.java
@@ -33,4 +33,4 @@ public class PartitionInfo {
   public LocalDateTime getCreateTime() {
     return createTime;
   }
-} 
+}
--- a/beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/hive/PartitionInfo.java
+++ b/beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/hive/PartitionInfo.java
@@ -1,0 +1,36 @@
+/**
+ * Copyright (C) 2019-2025 Expedia, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java
```diff
@@ -223,46 +224,53 @@ private void updateScheduledPartitions(HousekeepingMetadata metadata,
...
+    log.info("Scheduled {} partitions for table {}.{}", partitionInfo.size(), tableMetadata.getDatabaseName(),
         tableMetadata.getTableName());
   }
 
   private HousekeepingMetadata createNewMetadata(HousekeepingMetadata tableMetadata, String partitionName,
-      String path) {
+      PartitionInfo partitionInfo) {
+    LocalDateTime createTime = partitionInfo.getCreateTime();
+    
+    if (createTime == null) {
```
patduin, line 262: why don't you do this in the constructor of PartitionInfo so it will never return null. (Or alternatively throw an IllegalArgumentException if it is null). This should always be set so the calling class can just use it and we avoid extra checks.

beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/hive/PartitionInfo.java
```diff
@@ -0,0 +1,36 @@
...
+package com.expediagroup.beekeeper.scheduler.hive;
+
+import java.time.LocalDateTime;
+
+public class PartitionInfo {
+  private final String path;
+  private final LocalDateTime createTime;
+
+  public PartitionInfo(String path, LocalDateTime createTime) {
+    this.path = path;
```
patduin, line 25: Let's validate the arguments and throw IllegalArgument if null, it's up to the caller to set this so a valid value imo.

beekeeper-scheduler/src/main/java/com/expediagroup/beekeeper/scheduler/service/ExpiredHousekeepingMetadataSchedulerService.java
HamzaJugon: ok, moved the null check into partitionInfo and throw an illegalArgumentException.

Removed the null check here and use `partitionInfo.getCreateTime()` directly.

patduin (approved):

haroldjimenez:
Maybe you can add some integrations test to cover this scenario

haroldjimenez (approved):

